PR #5, opened 2020-08-09
Fix BGR rendering

DorianRudolph:
See https://github.com/alacritty/alacritty/issues/4100.

To my understanding, the issue is that freetype does not care about the subpixel order. It only differentiates between LCD and LCD_V and then generates a bitmap that either has threetimes the pixels in horizontal or vertical direction. We need to manually convert it into an RGB bitmap. Currently, this does not take into account the subpixel layout. 

Arguably, my code is somewhat inelegant, but I thought it would still be better to try to fix this myself rather than just complain...

### DIFF
--- a/src/ft/mod.rs
+++ b/src/ft/mod.rs
@@ -14,7 +14,7 @@ use log::{debug, trace};
 
 pub mod fc;
 
-use fc::{CharSet, FTFaceLocation, Pattern, PatternHash, PatternRef};
+use fc::{CharSet, FTFaceLocation, Pattern, PatternHash, PatternRef, Rgba};
 
 use super::{
     BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, Rasterize, RasterizedGlyph, Size, Slant,
@@ -55,7 +55,7 @@ struct FaceLoadingProperties {
     matrix: Option<Matrix>,
     pixelsize_fixup_factor: Option<f64>,
     ft_face: Rc<FTFace>,
-    rgba: fc::Rgba,
+    rgba: Rgba,
 }
 
 impl fmt::Debug for FaceLoadingProperties {
@@ -348,7 +348,7 @@ impl FreeTypeRasterizer {
 
             let pixelsize_fixup_factor = pattern.pixelsizefixupfactor().next();
 
-            let rgba = pattern.rgba().next().unwrap_or(fc::Rgba::Unknown);
+            let rgba = pattern.rgba().next().unwrap_or(Rgba::Unknown);
 
             let face = FaceLoadingProperties {
                 load_flags: Self::ft_load_flags(pattern),
@@ -498,7 +498,7 @@ impl FreeTypeRasterizer {
         let antialias = pattern.antialias().next().unwrap_or(true);
         let autohint = pattern.autohint().next().unwrap_or(false);
         let hinting = pattern.hinting().next().unwrap_or(true);
-        let rgba = pattern.rgba().next().unwrap_or(fc::Rgba::Unknown);
+        let rgba = pattern.rgba().next().unwrap_or(Rgba::Unknown);
         let embedded_bitmaps = pattern.embeddedbitmap().next().unwrap_or(true);
         let scalable = pattern.scalable().next().unwrap_or(true);
         let color = pattern.color().next().unwrap_or(false);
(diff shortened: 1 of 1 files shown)


DorianRudolph:
This is the equivalent code in cairo: 
 https://github.com/freedesktop/cairo/blob/1676935768b888b3e4038b6dddb4f6362c137345/src/cairo-ft-font.c#L1107

I suppose I could also do this directly in the `normalize_buffer` function. Would that be preferable to you?

chrisduerr:
> This is the equivalent code in cairo:
https://github.com/freedesktop/cairo/blob/1676935768b888b3e4038b6dddb4f6362c137345/src/cairo-ft-font.c#L1107

I'm not sure if it is. This seems to always do a lot of conversions even in cases where it shouldn't be necessary.

chrisduerr:
It seems like maybe something like https://www.freetype.org/freetype2/docs/reference/ft2-lcd_rendering.html#ft_library_setlcdgeometry could help. Though I'm not sure if that works by default.

DorianRudolph:
> This function does nothing and returns FT_Err_Unimplemented_Feature in the context of ClearType-style subpixel rendering when FT_CONFIG_OPTION_SUBPIXEL_RENDERING is defined in your build of the library.

Sounds to me like this might not work, depending on how the user's libfreetype is compiled. The freetype-sys create does not seem to expose the function.

Could you maybe elaborate what your performance concerns are? `normalize_buffer` already copies the bitmap pixel by pixel. I don't think changing the order there (as opposed to my original implementation as an additional pass) will impact performance measurably. I haven't profiled this, but I would suspect that even an additional pass should be negligible compared to time freetype takes for actually rendering the glyph, but I understand that you want as little overhead as possible.

chrisduerr:
> normalize_buffer already copies the bitmap pixel by pixel

Even more of a reason not to do it twice, especially in cases where it's just a no-op.

DorianRudolph:
Oh, my RGB->BGR iteration already only runs if necessary. 
```rust
        if face.reverse_rgb {
            if let BitmapBuffer::RGB(v) = &mut buf {
                for i in (0..v.len()).step_by(3) {
                    v.swap(i, i + 2);
                }
            }
        }
```

As previously mentioned, this could aslo be done directly in the normalize_buffer function:
```rust
            PixelMode::Lcd => {
                for i in 0..bitmap.rows() {
                    let start = (i as usize) * pitch;
                    let stop = start + bitmap.width() as usize;
                    if bgr {
                        for j in (start..stop).step_by(3) {
                            packed.push(buf[j + 2]);
                            packed.push(buf[j + 1]);
                            packed.push(buf[j]);
                        }
                    } else {
                        packed.extend_from_slice(&buf[start..stop]);
                    }
                }
                Ok((bitmap.rows(), bitmap.width() / 3, BitmapBuffer::RGB(packed)))
            },
            PixelMode::LcdV => {
                for i in 0..bitmap.rows() / 3 {
                    for j in 0..bitmap.width() {
                        for k in 0..3 {
                            let offset = ((i as usize) * 3 + if bgr {2 - k} else { k }) * pitch + (j as usize);
                            packed.push(buf[offset]);
                        }
                    }
                }    
                Ok((bitmap.rows() / 3, bitmap.width(), BitmapBuffer::RGB(packed)))
            },
```

chrisduerr:
> Oh, my RGB->BGR iteration already only runs if necessary (so it will never touch the data twice if not necessary).

It will touch the data twice, because the `normalize_buffer` function already touches it once.

DorianRudolph:
Yes, that is why I proposed the above modification to the normalize_buffer function. Then it will only iterate over the data once. Would that be an acceptable solution for you? If so, I would change my pull request to that.

chrisduerr:
It would certainly be preferable, yes. Though I'd have to check myself in which of the branches it's even necessary.

DorianRudolph:
Sure. I just pushed the changes.

src/ft/mod.rs
```diff
@@ -347,6 +348,11 @@ impl FreeTypeRasterizer {
 
             let pixelsize_fixup_factor = pattern.pixelsizefixupfactor().next();
 
+            let bgr = match pattern.rgba().next().unwrap_or(fc::Rgba::Unknown) {
+                fc::Rgba::Bgr | fc::Rgba::Vbgr => true,
+                _ => false,
+            };
```
kchibisov, line 354: Could we pass subpixel type around instead of bool flag, I don't like random bool flags, but on the other side other subpixel options won't be used?

What do you think @chrisduerr

chrisduerr: I completely agree. I also don't think it makes sense to match on an enum like that.

DorianRudolph:
Emoji work fine. According to the documentation `Bgra` is only used for color images, so it will not do any subpixel rendering.

There was a bug in the code where in the `LcdV` branch I matched againt `Bgr` and not `Vbgr` which I fixed.

To test different layouts, you can use in `~/.fonts.conf`.
```xml
 <match target="font">
  <edit mode="assign" name="rgba">
   <const>bgr</const>
  </edit>
 </match>
```

Here are screenshots of all 4 subpixel layouts:

RGB:
image

BGR:
image

VRGB:
image

VBGR:
image

For comparison, here are all four layouts.
image

chrisduerr (approved):
I've taken the liberty to extract `fc::Rgba` to a separate import, otherwise it's looking good to me.

Unfortunately I'm not able to properly verify font changes at the moment, since they're generally a bit fuzzy due to running a monitor at the incorrect resolution.

@kchibisov could you do a final visual review of this, just to make sure there aren't any rendering issues at least with RGB monitors. Zoomed in pictures looked good to me, I'd just like to double-check it.

kchibisov (approved):
Well, the code is the same for non rgb path, so it looks the same as you may think.

As for bgr it looks different, but I can't say much about it, since my monitor has rgb layout, but one thing that I can say that it's blurry compared to what it was, but if it was crisp on rgb display, then I guess it's expected?

Also, if you test such things you usually want to use `hintfull`(just saying).

chrisduerr:
> As for bgr it looks different, but I can't say much about it, since my monitor has rgb layout, but one thing that I can say that it's blurry compared to what it was, but if it was crisp on rgb display, then I guess it's expected?

Yes, that's exactly how it should be. Thanks for testing, LGTM.

PR #6, opened 2020-08-16
ft: Calculate height from bounding box

iamjamestl:
In alacritty/alacritty#2693, we see two problems: powerline separator glyphs are offset both vertically and horizontally.  Having meticulously redesigned the powerline separator glyphs for Hack in source-foundry/Hack#234, I can say with some degree of confidence that the horizontal misalignment is due to the separator glyphs not being designed to be larger than the bounding box.  That is to say, it's a font problem.

The vertical misaligment, however, comes down to how this library is calculating the font height.  FreeType provides a height metric, but it represents ["the vertical distance between two consecutive baselines"](https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_facerec) which doesn't necessarily correspond to the height of the glyphs.  Thus we get a situation where glyphs appear slightly above and below than the line on which their drawn.  You might want that for typesetting proportional fonts, but it doesn't make much sense in a monospace, terminal-like application.

For calculating the global glyph height, the FreeType documentation says to use the difference between the ascender and descender metrics.  Indeed, that is what [urxvt](https://github.com/exg/rxvt-unicode/blob/rxvt-unicode-rel-9.22/src/rxvtfont.C#L1244) and [xterm](https://github.com/Maximus5/xterm/blob/73ac00b39ec6f4af3b43908996a4bc73410a47df/src/fontutils.c#L833) do, and that is what is implemented in this patch.

This change may cause alacritty to render a pixel or two more space between lines depending on the font.  The spacing can be tweaked with the `font.offset` setting at the expense of powerline separator seamlessness.

alacritty-before-and-after-ft-height-change

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -19,11 +19,12 @@ tasks:
   - test: |
       cd crossfont
       cargo test
+  - oldstable: |
+      cd crossfont
+      rustup toolchain install --profile minimal 1.43.1
+      rustup default 1.43.1
+      cargo test
   - clippy: |
       cd crossfont
       rustup component add clippy
       cargo clippy --all-targets
-  - oldstable: |
-      cd crossfont
-      rustup toolchain install --profile minimal 1.43.1
-      cargo +1.43.1 test
--- a/.builds/linux.yml
+++ b/.builds/linux.yml
@@ -23,11 +23,12 @@ tasks:
       cd crossfont
       rustup toolchain install nightly -c rustfmt
       cargo +nightly fmt -- --check
+  - oldstable: |
+      cd crossfont
+      rustup toolchain install --profile minimal 1.43.1
+      rustup default 1.43.1
+      cargo test
   - clippy: |
       cd crossfont
       rustup component add clippy
       cargo clippy --all-targets
-  - oldstable: |
-      cd crossfont
-      rustup toolchain install --profile minimal 1.43.1
-      cargo +1.43.1 test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,11 +17,11 @@ jobs:
       - uses: actions/checkout@v2
       - name: Stable
         run: cargo test
-      - name: Clippy
-        run: |
-          rustup component add clippy
-          cargo clippy --all-targets
       - name: Oldstable
         run: |
           rustup default 1.43.1
           cargo test
+      - name: Clippy
+        run: |
+          rustup component add clippy
+          cargo clippy --all-targets
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- FreeType font height metric will now use `(ascent - descent)` if it is bigger than height
+
 ## 0.2.0
 
 ### Changed
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -416,7 +416,7 @@ impl Font {
                 height: 0,
                 top: 0,
                 left: 0,
-                buffer: BitmapBuffer::RGB(Vec::new()),
+                buffer: BitmapBuffer::Rgb(Vec::new()),
             };
         }
 
@@ -470,9 +470,9 @@ impl Font {
         let rasterized_pixels = cg_context.data().to_vec();
 
         let buffer = if is_colored {
-            BitmapBuffer::RGBA(byte_order::extract_rgba(&rasterized_pixels))
+            BitmapBuffer::Rgba(byte_order::extract_rgba(&rasterized_pixels))
         } else {
-            BitmapBuffer::RGB(byte_order::extract_rgb(&rasterized_pixels))
+            BitmapBuffer::Rgb(byte_order::extract_rgb(&rasterized_pixels))
         };
 
         RasterizedGlyph {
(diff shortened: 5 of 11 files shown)


iamjamestl:
> Why not? The height of a line should be based on the fonts line height, regardless of monospace or proportional? That's the point of the metric, isn't it?

I honestly don't know what the point of the height metric is because it's not accurate in most fonts and most apps don't use it (even this library uses some form of the ascender and descender metrics in calculating line height for [darwin](https://github.com/alacritty/crossfont/blob/fa0b7a960b3ea26bd98c5552c64d5847f5ffdbbd/src/darwin/mod.rs#L414) and [directwrite](https://github.com/alacritty/crossfont/blob/master/src/directwrite/mod.rs#L172)).  You can read all about that here: https://fontforge.org/docs/tutorial/editexample5.html#setting-the-baseline-to-baseline-spacing-of-a-font.

> > For calculating the global glyph height, the FreeType documentation says to use the difference between the ascender and descender metrics.
> 
> Please include a reference to the documentation.

I included a link in both the original description and the commit message, but here it is again: https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_facerec, and in quoted form:

> height: This value is the vertical distance between two consecutive  baselines, expressed in font units. It is always positive. Only relevant  for scalable formats. If you want the global glyph height, use `ascender - descender`.

iamjamestl:
> That doesn't really give any great insight about FreeType, so I don't think it's a good resource.

Sure, but it describes the reason why the height metric is unreliable, especially from a font designer's point of view--namely that the height metric is not available or not configurable in some font formats, but more importantly, that it's not really something to worry about because most renderers just use ascent +/- descent.

> In my experience the most relevant information about implementing these thing is what other applications do, since FreeType's docs are often a bit sparse. Looking at URxvt it seems to do both, just take whatever's bigger. I would assume that often fonts are probably bigger, rather than smaller. So that would still cause troubles. XTerm does use ascent + descent, but it's not really the pinnacle of font rendering, so I'm not sure how trustworthy it is here.

Ok, if what I view as **the** canonical X terminal emulators are not good enough, here are three more extremely popular ones:

1. **Konsole**

    Konsole calls `QFontMetrics::height` at https://github.com/KDE/konsole/blob/v20.08.0/src/TerminalDisplay.cpp#L229.  [`QFontMetrics::height`](https://doc.qt.io/archives/qt-4.8/qfontmetrics.html#height) is described as: 

    > Returns the height of the font. This is always equal to ascent()+descent()+1 (the 1 is for the base line).

    You can decide for yourself how the `+1` affects things, but it is not using the font's height metric.

2. **gnome-terminal/other VTE terms**

    Their only font backend, pango+cairo, calls `pango_layout_get_extents` at https://github.com/GNOME/vte/blob/09e71510d1b7061df177d6e3b531aa5cc86b8d65/src/fonts-pangocairo.cc#L186.

    `pango_layout_get_extents` goes through a couple of abstraction layers before calling `get_face_metrics` (https://github.com/GNOME/pango/blob/1ca61723a5256e3743e0e51c37923f417485cdc3/pango/pangofc-font.c#L328) which calculates the height metric thusly:

    ```
    metrics->height = extents.ascender - extents.descender + extents.line_gap;
    ```

    The addition of `extents.line_gap` is not too far off from what happens at https://github.com/alacritty/alacritty/blob/v0.5.0/alacritty/src/display.rs#L725 with the addition of `offset.y`.

3. **ST**

    ST gets the font height metric info from `xloadfont` at https://git.suckless.org/st/file/x.c.html#l989.  According to [xloadfont(3)](https://linux.die.net/man/3/xloadfont):

    > Its height is: ascent + descent

You mention Kitty, but it's clearly doing it's own sort of height fudging: the font size function tweaks the font size so the height metric matches the desired height.  At least that's how I'm reading this: https://github.com/kovidgoyal/kitty/blob/4062365b195284d5e13f17df3900dd6da252bbb7/kitty/freetype.c#L144.  For what it's worth, Kitty renders powerline glyphs correctly--still, I'd rather take the approach used by every other major terminal and their underlying font libraries than to fudge font size based on an unreliable metric.

So anyway, that's now five other implementations that use some combination of ascent and descent instead of the height metric.  That is backed by FreeType's own documentation, even if you think it is not very clear, and it's backed by sound reasoning from FontForge, which is probably the most popular open-source font design tool.

I approached this patch as someone who spent a lot of time redesigning powerline glyphs to look right in most terminals and other FreeType and ClearType applications.  I've put in the work.  Alacritty/crossfont is broken in this regard.  I don't think it's fair to say I'm asking for a change "just optimizing for one random font out there."  Alas, I think I have spent all the time on this that I want to so I'll leave it in your all's good hands.  Thanks for your consideration.

chrisduerr:
> Konsole calls QFontMetrics::height

So for them it's +1, that would be another option. The difference to others making it very unclear again what exact metrics most people really are using.

> Their only font backend, pango+cairo, calls pango_layout_get_extents

That is a good point, they seem to indeed be using asc - desc.

> ST gets the font height metric info from xloadfont at https://git.suckless.org/st/file/x.c.html#l989. According to xloadfont(3):

That's kinda pointless, since there's only ascent and descent. Of course if the library doesn't even allow you to access the height for whatever reason and the only data available is ascent + descent, those are recommended to calculate the height.

> That is backed by FreeType's own documentation, even if you think it is not very clear

That is not true. We want the distance between two baselines, not the global glyph height. What we care about is the height of a line, which should be the distance between two baselines. So just following freetype's documentation, this is the correct metric.

> it's backed by sound reasoning from FontForge

FontForge does not have any sound reasoning at all about Linux font rendering. So this is utterly useless as any kind of indicator.

So to summarize:
 - URxvt will use whatever's bigger
 - XTerm [seems to use whatever's bigger](https://github.com/Maximus5/xterm/blob/5d327b7f1aac55b623fe338801fe0e88e7ad99fe/src/fontutils.c#L2734)
 - Konsole uses ascent + descent + 1
 - VTE uses ascent + descent + line gap (this is important, because we do not have a default line gap)
 - Kitty uses height + shenanigans

While it might make sense to use ascent and descent in some capacity, I cannot say that these resources uniformly point at using just ascent + descent. In fact not a single one of them seems to do just that. So I'm not convinced that just using ascent + descent is sufficient here.

chrisduerr:
While height should work properly as metric for monospace fonts, I see nothing particularly wrong with using `ascent - descent` if it is bigger, especially with how many other applications use that. So I don't see a reason to hold this back unnecessarily.

chrisduerr (approved):

PR #7, opened 2020-08-27
Add license file

chrisduerr:


kchibisov:
I'd consider it licensing as MIT tbqh, however I'm not sure if we can just relicense it, what do you think?

chrisduerr:
> I'd consider it licensing as MIT tbqh, however I'm not sure if we can just relicense it, what do you think?

I think we absolutely cannot, no matter what we want. The original code was contributed under Apache, so we have no option but to stick with it. Unless you're interested in rewriting the entire library.

kchibisov:
>I think we absolutely cannot, no matter what we want. The original code was contributed under Apache, so we have no option but to stick with it. Unless you're interested in rewriting the entire library

I guess, unless we ask anyone that send patches about 'whether we can change the license'.

kchibisov (approved):

PR #8, opened 2020-09-23
Fix compilation with FreeType version below 2.8.0

kchibisov:
This should allow building on systems without recent FreeType
by excluding calls to recent FreeType features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,3 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Minimum Rust version has been bumped to 1.43.0
+
+### Fixed
+
+- Compilation with FreeType version below 2.8.0 on Linux/BSD
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,8 @@
+fn main() {
+    // This libtool version maps to FreeType version 2.8.0, so we can use
+    // `FT_Set_Default_Properties`.
+    #[cfg(not(any(target_os = "macos", windows)))]
+    if pkg_config::Config::new().atleast_version("20.0.14").probe("freetype2").is_ok() {
+        println!("cargo:rustc-cfg=ft_set_default_properties_available")
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,9 @@ log = "0.4"
 servo-fontconfig = "0.5.1"
 freetype-rs = "0.26"
 
+[target.'cfg(not(any(target_os = "macos", windows)))'.build-dependencies]
+pkg-config = "0.3"
+
 [target.'cfg(target_os = "macos")'.dependencies]
 cocoa = "0.20.1"
 core-foundation = "0.7"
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
chrisduerr, line 5:
```suggestion
# Changelog

All notable changes to Alacritty are documented in this file.
The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` and `Removed`.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
```

build.rs
```diff
@@ -0,0 +1,15 @@
+#[cfg(not(any(target_os = "macos", windows)))]
+extern crate pkg_config;
+
+#[cfg(not(any(target_os = "macos", windows)))]
+fn main() {
+    let library_version = pkg_config::probe_library("freetype2").unwrap().version;
```
chrisduerr, line 6: The unwrap here scares me. What does it do? I don't think it's the best idea.

Also can't we just use the following?

```rust
let enough = pkg_config::Config::new().atleast_version("2.8.0").probe("freetype2").is_ok();
```

While the `>=` vs `atleast_version` might not make a big difference, being able to specify the actual freetype version would.

chrisduerr (changes requested):

build.rs
kchibisov: Well, we can do something like that, I guess.

CHANGELOG.md
kchibisov: I'd assume that's crossfont in that case. Also, the header were from copypasta, so we should fix it as well?

chrisduerr: Yeah, sorry. I'm fine with copying it from copypasta too, yes. We probably just changed it in Alacritty after adding it to copypasta.

build.rs
chrisduerr: If that actually works, it would be much better I think.

kchibisov: I've updated and it works 'fine'.

build.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(not(any(target_os = "macos", windows)))]
+extern crate pkg_config;
+
+#[cfg(not(any(target_os = "macos", windows)))]
+fn main() {
+    if pkg_config::Config::new().atleast_version("20.0.14").probe("freetype2").is_ok() {
+        // This libtool version maps to 2.8.0 of freetype, so we can use
+        // `FT_Set_Default_Properties`.
```
chrisduerr, line 8: Put it above the conditional?

build.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(not(any(target_os = "macos", windows)))]
+extern crate pkg_config;
+
+#[cfg(not(any(target_os = "macos", windows)))]
+fn main() {
+    if pkg_config::Config::new().atleast_version("20.0.14").probe("freetype2").is_ok() {
```
chrisduerr, line 6:
```suggestion
fn main() {
    #[cfg(not(any(target_os = "macos", windows)))]
    if pkg_config::Config::new().atleast_version("20.0.14").probe("freetype2").is_ok() {
```

I think this would be much easier, with the new annotations support?

chrisduerr (changes requested):

chrisduerr (approved):

PR #9, opened 2020-09-23
Bump minimum rust version to 1.43

kchibisov:


chrisduerr (changes requested):
Want to add the changelog in this one?

kchibisov:
I'm not sure that it matters?

kchibisov:
Note, there's no such thing as a minimum rust version on a cargo toml level. it matters for alacritty, but I'm not sure that it matters for us.

chrisduerr:
> Note, there's no such thing as a minimum rust version on a cargo toml level. it matters for alacritty, but I'm not sure that it matters for us.

It does. Because clients like Alacritty will fail building if the minimum version is bumped without clear communication.

kchibisov:
>It does. Because clients like Alacritty will fail building if the minimum version is bumped without clear communication.

In which section that change should go?

chrisduerr (approved):

PR #10, opened 2020-10-16
Bump linux and macos dependencies

returntrip:


chrisduerr (commented):
Is there a specific reason why you updated just euclid?

returntrip:
I am trying to package alacritty in the official Fedora repo: https://bugzilla.redhat.com/show_bug.cgi?id=1888923.
Fedora does not vendor in crates and they are built from scratch so we try to keep all the crates/deps to the latest version.
Euclid is at v 0.22.1 in our repo (https://src.fedoraproject.org/rpms/rust-euclid) hence I am bumping it here.

kchibisov:
>Is there a specific reason why you updated just euclid?

No, the question was 'why only euclid was updated', we have other deps as well here.

Would you mind bumping macOS deps as well, they shouldn't really be breaking.

chrisduerr:
> Would you mind bumping macOS deps as well, they shouldn't really be breaking.

Yeah exactly that was my main question. Though of course "I don't care about those others" is an explanation for that. But I'd still like to upgrade the other dependencies of course.

returntrip:
Sorry folks, hope this PR is correct now. There was no intention to ignore macos, mine is a lack of knowledge, the less I change the less I can break.

chrisduerr (approved):
Thank you!

kchibisov:
>I am trying to package alacritty in the official Fedora repo: https://bugzilla.redhat.com/show_bug.cgi?id=1888923.
Fedora does not vendor in crates and they are built from scratch so we try to keep all the crates/deps to the latest version.

Btw, there's a change that 0.6.0 version of alacritty won't have any duplicates at all, since the only one that left is in https://github.com/expenses/line_drawing and @chrisduerr sent a patch for it already https://github.com/expenses/line_drawing/pull/17

PR #11, opened 2020-10-16
Fix clippy warnings

chrisduerr:


src/darwin/mod.rs
```diff
@@ -192,14 +192,8 @@ impl Rasterizer {
...
-        let bold = match weight {
-            Weight::Bold => true,
-            _ => false,
-        };
-        let italic = match slant {
-            Slant::Normal => false,
-            _ => true,
-        };
+        let bold = weight == Weight::Bold;
+        let italic = slant != Slant::Normal;
```
kchibisov, line 196: we can use matches macro since it's from 1.42, and we're using 1.43.

chrisduerr: Why would we? This is better than the matches macro.

kchibisov (approved):

PR #15, opened 2020-11-09
Return missing glyph as Error

kchibisov:
This should help users identify when they failed to
get the requested glyph, since it was missing on the system.
The resulted missing glyph will be stored in 'Error::MissingGlyph',
so they can easily render it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Rasterizer no longer uses associated platform-specific error type and now has common type for `Error`
-- If the glyph you've requested is missing, the resulting missing glyph char will be in `Error::MissingGlyph`
+- The rasterizer's `Error` type is now shared across platforms
+- Missing glyphs are now returned as the content of the `MissingGlyph` error
 - `RasterizedGlyph`'s `c` and `buf` fields are now named `character` and `buffer` respectively
 - `GlyphKey`'s `c` field is now named `character`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- Rasterizer no longer uses associated `Err` type and now has common type for `Error`
+- If the glyph you've requested is missing, the resulted missing glyph char will be in `Error::MissingGlyph`
+
 ## 0.1.1
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Rasterizer no longer uses associated `Err` type and now has common type for `Error`
-- If the glyph you've requested is missing, the resulted missing glyph char will be in `Error::MissingGlyph`
+- Rasterizer no longer uses associated platform-specific error type and now has common type for `Error`
+- If the glyph you've requested is missing, the resulting missing glyph char will be in `Error::MissingGlyph`
+- `RasterizedGlyph`'s `c` and `buf` fields are now named `character` and `buffer` respectively
+- `GlyphKey`'s `c` field is now named `character`
 
 ## 0.1.1
 
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -2,6 +2,7 @@
 
 #![allow(improper_ctypes)]
 use std::collections::HashMap;
+use std::iter;
 use std::path::PathBuf;
 use std::ptr;
 
@@ -89,7 +90,7 @@ impl crate::Rasterize for Rasterizer {
 
     /// Get metrics for font specified by FontKey.
     fn metrics(&self, key: FontKey, _size: Size) -> Result<Metrics, Error> {
-        let font = self.fonts.get(&key).ok_or(Error::FontNotLoaded)?;
+        let font = self.fonts.get(&key).ok_or(Error::UnknownFontKey)?;
 
         Ok(font.metrics())
     }
@@ -110,30 +111,24 @@ impl crate::Rasterize for Rasterizer {
     /// Get rasterized glyph for given glyph key.
     fn get_glyph(&mut self, glyph: GlyphKey) -> Result<RasterizedGlyph, Error> {
         // Get loaded font.
-        let font = self.fonts.get(&glyph.font_key).ok_or(Error::FontNotLoaded)?;
+        let font = self.fonts.get(&glyph.font_key).ok_or(Error::UnknownFontKey)?;
 
-        // Find a font where the given char is presented.
-        let mut glyph_index = font.glyph_index(glyph.c);
-        let font = match glyph_index {
-            Some(_) => font,
-            None => font
-                .fallbacks
-                .iter()
-                .find(|font| {
-                    glyph_index = font.glyph_index(glyph.c);
-                    glyph_index.is_some()
-                })
-                .unwrap_or(font),
-        };
+        // Find a font where the given char is present.
+        let (font, glyph_index) = iter::once(font)
+            .chain(font.fallbacks.iter())
+            .find_map(|font| {
+                let glyph_index = font.glyph_index(glyph.character);
+                glyph_index.map(|glyph_index| (font, Some(glyph_index)))
+            })
+            .unwrap_or((font, None));
 
         let is_missing_glyph = glyph_index.is_none();
 
-        // According to
+        // According to the index of 0 must be a missing glyph character:
         // https://developer.apple.com/fonts/TrueType-Reference-Manual/RM07/appendixB.html
-        // the index of 0 must be a missing glyph character.
         let glyph_index = glyph_index.unwrap_or(0);
 
-        let glyph = font.get_glyph(glyph.c, glyph_index, self.use_thin_strokes);
+        let glyph = font.get_glyph(glyph.character, glyph_index, self.use_thin_strokes);
 
         if is_missing_glyph {
             Err(Error::MissingGlyph(glyph))
@@ -164,7 +159,7 @@ impl Rasterizer {
             }
         }
 
-        Err(Error::MissingFont(desc.to_owned()))
+        Err(Error::FontNotFound(desc.to_owned()))
     }
 
     fn get_matching_face(
@@ -187,7 +182,7 @@ impl Rasterizer {
             }
         }
 
-        Err(Error::MissingFont(desc.to_owned()))
+        Err(Error::FontNotFound(desc.to_owned()))
     }
 
     fn get_font(&mut self, desc: &FontDesc, size: Size) -> Result<Font, Error> {
@@ -436,12 +431,12 @@ impl Font {
 
         if rasterized_width == 0 || rasterized_height == 0 {
             return RasterizedGlyph {
-                c: ' ',
+                character: ' ',
                 width: 0,
                 height: 0,
                 top: 0,
                 left: 0,
-                buf: BitmapBuffer::RGB(Vec::new()),
+                buffer: BitmapBuffer::RGB(Vec::new()),
             };
         }
 
@@ -494,26 +489,26 @@ impl Font {
 
         let rasterized_pixels = cg_context.data().to_vec();
 
-        let buf = if is_colored {
+        let buffer = if is_colored {
             BitmapBuffer::RGBA(byte_order::extract_rgba(&rasterized_pixels))
         } else {
             BitmapBuffer::RGB(byte_order::extract_rgb(&rasterized_pixels))
         };
 
         RasterizedGlyph {
-            c: character,
+            character,
             left: rasterized_left,
             top: (bounds.size.height + bounds.origin.y).ceil() as i32,
             width: rasterized_width as i32,
             height: rasterized_height as i32,
-            buf,
+            buffer,
         }
     }
 
     fn glyph_index(&self, character: char) -> Option<u32> {
         // Encode this char as utf-16.
-        let mut buf = [0; 2];
-        let encoded: &[u16] = character.encode_utf16(&mut buf);
+        let mut buffer = [0; 2];
+        let encoded: &[u16] = character.encode_utf16(&mut buffer);
         // And use the utf-16 buffer to get the index.
         self.glyph_index_utf16(encoded)
     }
@@ -566,16 +561,15 @@ mod tests {
                 let glyph_index = font.glyph_index(*character).unwrap();
                 let glyph = font.get_glyph(*character, glyph_index, false);
 
-                let buf = match &glyph.buf {
-                    BitmapBuffer::RGB(buf) => buf,
-                    BitmapBuffer::RGBA(buf) => buf,
+                let buffer = match &glyph.buffer {
+                    BitmapBuffer::RGB(buffer) | BitmapBuffer::RGBA(buffer) => buffer,
                 };
 
                 // Debug the glyph.. sigh.
                 for row in 0..glyph.height {
                     for col in 0..glyph.width {
                         let index = ((glyph.width * 3 * row) + (col * 3)) as usize;
-                        let value = buf[index];
+                        let value = buffer[index];
                         let c = match value {
                             0..=50 => ' ',
                             51..=100 => '.',
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -36,7 +36,8 @@ pub mod byte_order;
 use byte_order::kCGBitmapByteOrder32Host;
 
 use super::{
-    BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style, Weight,
+    BitmapBuffer, Error, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style,
+    Weight,
 };
 
 /// Font descriptor.
@@ -76,42 +77,7 @@ pub struct Rasterizer {
     use_thin_strokes: bool,
 }
 
-/// Errors occurring when using the core text rasterizer.
-#[derive(Debug)]
-pub enum Error {
-    /// Tried to rasterize a glyph but it was not available.
-    MissingGlyph(char),
-
-    /// Couldn't find font matching description.
-    MissingFont(FontDesc),
-
-    /// Requested an operation with a FontKey that isn't known to the rasterizer.
-    FontNotLoaded,
-}
-
-impl ::std::error::Error for Error {
-    fn description(&self) -> &str {
-        match *self {
-            Error::MissingGlyph(ref _c) => "Unable to find the requested glyph",
-            Error::MissingFont(ref _desc) => "Unable to find the requested font",
-            Error::FontNotLoaded => "Tried to operate on font that hasn't been loaded",
-        }
-    }
-}
-
-impl ::std::fmt::Display for Error {
-    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-        match *self {
-            Error::MissingGlyph(ref c) => write!(f, "Glyph not found for char {:?}", c),
-            Error::MissingFont(ref desc) => write!(f, "Unable to find the font {}", desc),
-            Error::FontNotLoaded => f.write_str("Tried to use a font that hasn't been loaded"),
-        }
-    }
-}
-
 impl crate::Rasterize for Rasterizer {
-    type Err = Error;
-
     fn new(device_pixel_ratio: f32, use_thin_strokes: bool) -> Result<Rasterizer, Error> {
         Ok(Rasterizer {
             fonts: HashMap::new(),
@@ -146,18 +112,34 @@ impl crate::Rasterize for Rasterizer {
         // Get loaded font.
         let font = self.fonts.get(&glyph.font_key).ok_or(Error::FontNotLoaded)?;
 
-        // First try the font itself as a direct hit.
-        self.maybe_get_glyph(glyph, font).unwrap_or_else(|| {
-            // Then try fallbacks.
-            for fallback in &font.fallbacks {
-                if let Some(result) = self.maybe_get_glyph(glyph, &fallback) {
-                    // Found a fallback.
-                    return result;
-                }
-            }
-            // No fallback, give up.
-            Err(Error::MissingGlyph(glyph.c))
-        })
+        // Find a font where the given char is presented.
+        let mut glyph_index = font.glyph_index(glyph.c);
+        let font = match glyph_index {
+            Some(_) => font,
+            None => font
+                .fallbacks
+                .iter()
+                .find(|font| {
+                    glyph_index = font.glyph_index(glyph.c);
+                    glyph_index.is_some()
+                })
+                .unwrap_or(font),
+        };
+
+        let is_missing_glyph = glyph_index.is_none();
+
+        // According to
+        // https://developer.apple.com/fonts/TrueType-Reference-Manual/RM07/appendixB.html
+        // the index of 0 must be a missing glyph character.
+        let glyph_index = glyph_index.unwrap_or(0);
+
+        let glyph = font.get_glyph(glyph.c, glyph_index, self.use_thin_strokes);
+
+        if is_missing_glyph {
+            Err(Error::MissingGlyph(glyph))
+        } else {
+            Ok(glyph)
+        }
     }
 
     fn update_dpr(&mut self, device_pixel_ratio: f32) {
@@ -215,21 +197,6 @@ impl Rasterizer {
                 self.get_matching_face(desc, slant, weight, size)
             },
         }
-    }
-
-    /// Helper to try and get a glyph for a given font. Used for font fallback.
-    fn maybe_get_glyph(
-        &self,
-        glyph: GlyphKey,
-        font: &Font,
-    ) -> Option<Result<RasterizedGlyph, Error>> {
-        let scaled_size = self.device_pixel_ratio * glyph.size.as_f32_pts();
-        font.get_glyph(glyph.c, f64::from(scaled_size), self.use_thin_strokes)
-            .map(|r| Some(Ok(r)))
-            .unwrap_or_else(|e| match e {
-                Error::MissingGlyph(_) => None,
-                _ => Some(Err(e)),
-            })
     }
 }
 
@@ -455,12 +422,9 @@ impl Font {
     pub fn get_glyph(
         &self,
         character: char,
-        _size: f64,
+        glyph_index: u32,
         use_thin_strokes: bool,
-    ) -> Result<RasterizedGlyph, Error> {
-        let glyph_index =
-            self.glyph_index(character).ok_or_else(|| Error::MissingGlyph(character))?;
-
+    ) -> RasterizedGlyph {
         let bounds = self.bounding_rect_for_glyph(Default::default(), glyph_index);
 
         let rasterized_left = bounds.origin.x.floor() as i32;
@@ -471,14 +435,14 @@ impl Font {
         let rasterized_height = (rasterized_descent + rasterized_ascent) as u32;
 
         if rasterized_width == 0 || rasterized_height == 0 {
-            return Ok(RasterizedGlyph {
+            return RasterizedGlyph {
                 c: ' ',
                 width: 0,
                 height: 0,
                 top: 0,
                 left: 0,
                 buf: BitmapBuffer::RGB(Vec::new()),
-            });
+            };
         }
 
         let mut cg_context = CGContext::create_bitmap_context(
@@ -536,14 +500,14 @@ impl Font {
             BitmapBuffer::RGB(byte_order::extract_rgb(&rasterized_pixels))
         };
 
-        Ok(RasterizedGlyph {
+        RasterizedGlyph {
             c: character,
             left: rasterized_left,
             top: (bounds.size.height + bounds.origin.y).ceil() as i32,
             width: rasterized_width as i32,
             height: rasterized_height as i32,
             buf,
-        })
+        }
     }
 
     fn glyph_index(&self, character: char) -> Option<u32> {
@@ -598,8 +562,9 @@ mod tests {
 
         for font in fonts {
             // Get a glyph.
-            for c in &['a', 'b', 'c', 'd'] {
-                let glyph = font.get_glyph(*c, 72., false).unwrap();
+            for character in &['a', 'b', 'c', 'd'] {
+                let glyph_index = font.glyph_index(*character).unwrap();
+                let glyph = font.get_glyph(*character, glyph_index, false);
 
                 let buf = match &glyph.buf {
                     BitmapBuffer::RGB(buf) => buf,
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -41,6 +41,10 @@ use super::{
     Weight,
 };
 
+/// According to the documentation, the index of 0 must be a missing glyph character:
+/// https://developer.apple.com/fonts/TrueType-Reference-Manual/RM07/appendixB.html
+const MISSING_GLYPH_INDEX: u32 = 0;
+
 /// Font descriptor.
 ///
 /// The descriptor provides data about a font and supports creating a font.
@@ -116,21 +120,15 @@ impl crate::Rasterize for Rasterizer {
         // Find a font where the given char is present.
         let (font, glyph_index) = iter::once(font)
             .chain(font.fallbacks.iter())
-            .find_map(|font| {
-                let glyph_index = font.glyph_index(glyph.character);
-                glyph_index.map(|glyph_index| (font, Some(glyph_index)))
+            .find_map(|font| match font.glyph_index(glyph.character) {
+                MISSING_GLYPH_INDEX => None,
+                glyph_index => Some((font, glyph_index)),
             })
-            .unwrap_or((font, None));
-
-        let is_missing_glyph = glyph_index.is_none();
-
-        // According to the index of 0 must be a missing glyph character:
-        // https://developer.apple.com/fonts/TrueType-Reference-Manual/RM07/appendixB.html
-        let glyph_index = glyph_index.unwrap_or(0);
+            .unwrap_or((font, MISSING_GLYPH_INDEX));
 
         let glyph = font.get_glyph(glyph.character, glyph_index, self.use_thin_strokes);
 
-        if is_missing_glyph {
+        if glyph_index == MISSING_GLYPH_INDEX {
             Err(Error::MissingGlyph(glyph))
         } else {
             Ok(glyph)
@@ -400,7 +398,7 @@ impl Font {
     }
 
     fn glyph_advance(&self, character: char) -> f64 {
-        let index = self.glyph_index(character).unwrap();
+        let index = self.glyph_index(character);
 
         let indices = [index as CGGlyph];
 
@@ -505,7 +503,7 @@ impl Font {
         }
     }
 
-    fn glyph_index(&self, character: char) -> Option<u32> {
+    fn glyph_index(&self, character: char) -> u32 {
         // Encode this char as utf-16.
         let mut buffer = [0; 2];
         let encoded: &[u16] = character.encode_utf16(&mut buffer);
@@ -513,7 +511,7 @@ impl Font {
         self.glyph_index_utf16(encoded)
     }
 
-    fn glyph_index_utf16(&self, encoded: &[u16]) -> Option<u32> {
+    fn glyph_index_utf16(&self, encoded: &[u16]) -> u32 {
         // Output buffer for the glyph. for non-BMP glyphs, like
         // emojis, this will be filled with two chars the second
         // always being a 0.
@@ -528,9 +526,9 @@ impl Font {
         };
 
         if res {
-            Some(u32::from(glyphs[0]))
+            u32::from(glyphs[0])
         } else {
-            None
+            MISSING_GLYPH_INDEX
         }
     }
 }
@@ -558,7 +556,7 @@ mod tests {
         for font in fonts {
             // Get a glyph.
             for character in &['a', 'b', 'c', 'd'] {
-                let glyph_index = font.glyph_index(*character).unwrap();
+                let glyph_index = font.glyph_index(*character);
                 let glyph = font.get_glyph(*character, glyph_index, false);
 
                 let buffer = match &glyph.buffer {
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -117,7 +117,7 @@ impl crate::Rasterize for Rasterizer {
         // Get loaded font.
         let font = self.fonts.get(&glyph.font_key).ok_or(Error::UnknownFontKey)?;
 
-        // Find a font where the given char is present.
+        // Find a font where the given character is present.
         let (font, glyph_index) = iter::once(font)
             .chain(font.fallbacks.iter())
             .find_map(|font| match font.glyph_index(glyph.character) {
--- a/src/directwrite/mod.rs
+++ b/src/directwrite/mod.rs
@@ -3,7 +3,6 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::ffi::OsString;
-use std::iter;
 use std::os::windows::ffi::OsStringExt;
 
 use dwrote::{
@@ -19,6 +18,10 @@ use super::{
     BitmapBuffer, Error, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style,
     Weight,
 };
+
+/// DirectWrite uses 0 for missing glyph symbols.
+/// https://docs.microsoft.com/en-us/typography/opentype/spec/recom#glyph-0-the-notdef-glyph
+const MISSING_GLYPH_INDEX: u16 = 0;
 
 /// Cached DirectWrite font.
 struct Font {
@@ -95,12 +98,9 @@ impl DirectWriteRasterizer {
         self.fonts.get(&font_key).ok_or(Error::UnknownFontKey)
     }
 
-    fn get_glyph_index(&self, face: &FontFace, character: char) -> Option<u16> {
+    fn get_glyph_index(&self, face: &FontFace, character: char) -> u16 {
         // DirectWrite returns 0 if the glyph does not exist in the font.
-        face.get_glyph_indices(&[character as u32])
-            .first()
-            .copied()
-            .filter(|glyph_index| glyph_index != &0)
+        face.get_glyph_indices(&[character as u32]).first().copied().unwrap_or(MISSING_GLYPH_INDEX)
     }
 
     fn get_fallback_font(&self, loaded_font: &Font, character: char) -> Option<dwrote::Font> {
@@ -166,7 +166,7 @@ impl crate::Rasterize for DirectWriteRasterizer {
 
         // Since all monospace characters have the same width, we use `!` for horizontal metrics.
         let character = '!';
-        let glyph_index = self.get_glyph_index(face, character).unwrap_or(0);
+        let glyph_index = self.get_glyph_index(face, character);
 
         let glyph_metrics = face.get_design_glyph_metrics(&[glyph_index], false);
         let hmetrics = glyph_metrics.first().ok_or(Error::MissingSizeMetrics)?;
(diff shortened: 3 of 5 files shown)


kchibisov:
I've tested this on macOS and now I actually see missing glyph symbols.

src/darwin/mod.rs
```diff
@@ -146,18 +112,34 @@ impl crate::Rasterize for Rasterizer {
...
+                    glyph_index.is_some()
+                })
+                .unwrap_or(font),
+        };
+
+        let is_missing_glyph = glyph_index.is_none();
+
+        // According to
+        // https://developer.apple.com/fonts/TrueType-Reference-Manual/RM07/appendixB.html
+        // the index of 0 must be a missing glyph character.
```
chrisduerr, line 133:
```suggestion
        // The index of `0` always corresponds to the missing glyph character:
        // https://developer.apple.com/fonts/TrueType-Reference-Manual/RM07/appendixB.html
```

Just easier to say everything up front, then formatting doesn't get messed up by the URL.

src/darwin/mod.rs
```diff
@@ -216,21 +198,6 @@ impl Rasterizer {
...
         }
     }
-
-    /// Helper to try and get a glyph for a given font. Used for font fallback.
-    fn maybe_get_glyph(
-        &self,
-        glyph: GlyphKey,
-        font: &Font,
-    ) -> Option<Result<RasterizedGlyph, Error>> {
-        let scaled_size = self.device_pixel_ratio * glyph.size.as_f32_pts();
```
chrisduerr, line 226: Huh I was kinda curious where this line has gone... but looks like it was never used for anything in the first place?

src/directwrite/mod.rs
```diff
@@ -73,18 +72,18 @@ impl DirectWriteRasterizer {
             0.0,
             0.0,
         )
-        .map_err(Error::DirectWriteError)?;
+        .map_err(Error::from)?;
```
chrisduerr, line 75: Shouldn't `Into` be automatically invoked by `?`?

src/directwrite/mod.rs
```diff
@@ -73,18 +72,18 @@ impl DirectWriteRasterizer {
...
+            .map_err(Error::from)?;
 
         let buf = glyph_analysis
             .create_alpha_texture(dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1, bounds)
-            .map_err(Error::DirectWriteError)?;
+            .map_err(Error::from)?;
 
         Ok(RasterizedGlyph {
-            c,
+            c: character,
```
chrisduerr, line 86: I'm not sure, but if you rename things anyways, it might be good to do it uniformly anywhere including the `RasterizedGlyph`?

src/directwrite/mod.rs
```diff
@@ -97,15 +96,12 @@ impl DirectWriteRasterizer {
         self.fonts.get(&font_key).ok_or(Error::FontNotLoaded)
     }
 
-    fn get_glyph_index(&self, face: &FontFace, c: char) -> Result<u16, Error> {
-        let idx = *face
-            .get_glyph_indices(&[c as u32])
-            .first()
+    fn get_glyph_index(&self, face: &FontFace, c: char) -> Option<u16> {
```
chrisduerr, line 99: Another instance of inconsistent `c` if your goal was to make that more consistent across backends.

src/directwrite/mod.rs
```diff
@@ -97,15 +96,12 @@ impl DirectWriteRasterizer {
...
+        match *face.get_glyph_indices(&[c as u32]).first()? {
```
chrisduerr, line 100: Why not stick to `.filter`? That seems like the easiest solution here. If you return an Option anyways, you can easily filter that out, right?

src/directwrite/mod.rs
```diff
@@ -172,11 +166,11 @@ impl crate::Rasterize for DirectWriteRasterizer {
         let line_height = f64::from(ascent - descent + line_gap);
 
         // Since all monospace characters have the same width, we use `!` for horizontal metrics.
-        let c = '!';
-        let glyph_index = self.get_glyph_index(face, c)?;
+        let character = '!';
+        let glyph_index = self.get_glyph_index(face, character).unwrap_or(0);
```
chrisduerr, line 170: DirectWrite returns `0` if the glyph cannot be found in the font, but does that mean that `0` is a safe glyph index to use ourselves?

If so, does it still return the correct metrics and could we just always use that?

src/directwrite/mod.rs
```diff
@@ -172,11 +166,11 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
 
         let glyph_metrics = face.get_design_glyph_metrics(&[glyph_index], false);
-        let hmetrics = glyph_metrics.first().ok_or_else(|| Error::MissingGlyph(c))?;
+        let hmetrics = glyph_metrics.first().ok_or(Error::MissingSizeMetrics)?;
```
chrisduerr, line 173: I'm not sure if `MissingSizeMetrics` is the right call here, does this error only happen when the glyph index doesn't exist or is there another case?

Windows docs don't say anything, so I'm fine with "Don't know, so I'm just being cautious with a dedicated error type".

src/directwrite/mod.rs
```diff
@@ -238,12 +232,33 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
+                None => loaded_font,
+                Some(font) => {
+                    fallback_font = Font::from(font);
+                    glyph_index = self.get_glyph_index(&fallback_font.face, glyph.c);
+
+                    &fallback_font
+                },
             },
-            result => result,
+        };
```
chrisduerr, line 249: This seems like it would be something that could also be cleaned up a bit. If the other suggestion doesn't help with doing so please let me know and I'll have to play around with it myself.

src/directwrite/mod.rs
```diff
@@ -333,3 +325,10 @@ impl TextAnalysisSourceMethods for TextAnalysisSourceData<'_> {
         dwrite::DWRITE_READING_DIRECTION_LEFT_TO_RIGHT
     }
 }
+
+impl From<HRESULT> for Error {
+    fn from(hresult: HRESULT) -> Self {
+        let message = format!("A DirectWrite rendering error occurred: {:#X}", hresult);
+        Error::PlatformError(message)
+    }
+}
```
chrisduerr, line 334: I feel like ideally we'd return proper errors here, but I can see why that would not be useful for downstream and unnecessarily complicated for crossfont, so it's probably fine.

src/ft/mod.rs
```diff
@@ -481,17 +480,22 @@ impl FreeTypeRasterizer {
...
-                pixelsize_fixup_factor
-            } else {
-                // Fallback if user has bitmap scaling disabled.
-                let metrics = face.ft_face.size_metrics().ok_or(Error::MissingSizeMetrics)?;
-                f64::from(pixelsize) / f64::from(metrics.y_ppem)
-            };
-            Ok(downsample_bitmap(rasterized_glyph, fixup_factor))
-        } else {
-            Ok(rasterized_glyph)
+        match (is_missing_glyph, face.colored) {
```
chrisduerr, line 483: My mash these together when all you do for a missing glyph is just return early anyways? That seems unnecessarily complicated, wouldn't it be cleaner to just have a separate check for that which returns right before this?

Or at least have an `if/elseif/else` block, rather than a 'match everything at the same time' approach. That's usually easier to follow and probably the best idea here when we're dealing only with booleans anyways.

src/lib.rs
```diff
@@ -203,23 +203,58 @@ pub struct Metrics {
...
 }
 
-pub trait Rasterize {
-    /// Errors occurring in Rasterize methods.
-    type Err: ::std::error::Error + Send + Sync + 'static;
+/// Errors occuring when using the rasterizer.
+#[derive(Debug)]
+pub enum Error {
+    /// Couldn't find font matching description.
+    MissingFont(FontDesc),
```
chrisduerr, line 210: I'm not sure if "Missing" is the correct word here? Maybe something like `FontNotFound`, akin to `std::io::ErrorKind::NotFound`?

src/lib.rs
```diff
@@ -203,23 +203,58 @@ pub struct Metrics {
...
+
+    /// Tried to get size metrics from a Face that didn't have a size.
+    MissingSizeMetrics,
```
chrisduerr, line 213: What does "missing" even mean here? Is that actually the error, that the metrics are missing for the font/glyph? I don't think that's the case.

Also not sure what anyone is supposed to do with this error, over just a generic `PlatformError`? Seems like there's no way to handle it anyways and it should basically never happen?

src/lib.rs
```diff
@@ -203,23 +203,58 @@ pub struct Metrics {
...
+
+    MissingGlyph(RasterizedGlyph),
```
chrisduerr, line 215: Could probably also remove the missing here too, but it's the most appropriate place probably. But even that is questionable.

src/lib.rs
```diff
@@ -203,23 +203,58 @@ pub struct Metrics {
...
+
+    /// Requested an operation with a FontKey that isn't known to the rasterizer.
+    FontNotLoaded,
```
chrisduerr, line 218: Instead of saying that the font isn't loaded, shouldn't we say that the glyph key isn't valid? Since that's how the enduser interacts with our API.

Since it's `get_glyph(GLYPH_KEY)`, I wouldn't expect the error to tell me that the font isn't loaded, but just that the glyph key is invalid. For whatever reason that is, is up to me?

src/lib.rs
```diff
@@ -203,23 +203,58 @@ pub struct Metrics {
...
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::MissingFont(font) => write!(f, "Unable to find the font {}", font),
```
chrisduerr, line 233:
```suggestion
            Error::MissingFont(font) => write!(f, "Unable to find the font {:?}", font),
```

To quote it, might make it easier to understand? Could solve potential issues like searching for `Fira Mono<lots of whitespace>`.

src/darwin/mod.rs
kchibisov: Yeah, it was even silenced before. just strange...

src/directwrite/mod.rs
kchibisov: It should be safe, since it's more about font design, thus the first index in a font should be 0 across the platforms. Missing glyph is also a glyph, so it's fine to use it.

src/directwrite/mod.rs
kchibisov: Well, I've assumed that if you can't get metrics it's a metrics issue.

src/directwrite/mod.rs
kchibisov: The problem is that you should create a font in some cases, which is not the same as on macOS with similar construction, though might work.

src/directwrite/mod.rs
kchibisov: I was thinking about having PlatformError(SomeErrorType), but it seemed to me that there's literally no benefit in it and the only thing that matters is the message, so I've preserved the message and users will still know it, just removed complexity from using nested errors.

src/lib.rs
kchibisov: I think I've ported it from our FreeType backend, I think it's what they are doing though. Well, there's a way to handle it, to request other font.

src/directwrite/mod.rs
chrisduerr: > thus the first index in a font should be 0 across the platforms

That's completely implementation dependent as far as I'm aware. And this code is exclusively about directwrite anyways.

src/directwrite/mod.rs
chrisduerr: Well `Missing` indicates a specific metrics issue though.

src/lib.rs
kchibisov: Well, I'd like to keep `MissingGlyph` for this one, since the concept it wraps is called `MissingGlyph`.

src/lib.rs
kchibisov: It seems like all the underlying API just returns `Option` for missing metrics, and I feel like you can load a font, but which has something strange with metrics going on, so I'd keep that error for now...

src/directwrite/mod.rs
kchibisov: Yeah yeah, we try to get metrics one line above, then check if we actually got, if not it's `Missing`, seems fine to me?

chrisduerr: The error isn't that it's missing though, the error is that we couldn't retrieve it. Why we weren't able to do so we don't know.

Missing metrics would indicate that the font doesn't have any metrics to be retrieved, but we don't know that.

src/lib.rs
chrisduerr: Not sure what you're talking about. You mean the other APIs (linux/macOS) return `Option`?

kchibisov: Yeah, there's no error to map, you can generate something though.

kchibisov:
I've updated everything except for naming `MissingSizeMetrics`, since it's still not clear to me with what to replace.

src/directwrite/mod.rs
kchibisov: @davidhewitt  Do you happen to know how this `0` index story works on Windows? I can't really find anything related in docs wrt that, I just know that it 0 mean undefined according to https://docs.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritefontface-getglyphindices . I think it should be safe to use 0 tbqh, though I can't really test it.

davidhewitt: Looks like there's some info here: https://docs.microsoft.com/en-us/typography/opentype/spec/recom#glyph-0-the-notdef-glyph

I can give this branch a test-drive on Windows if you tell me what I should look out for?

CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- Rasterizer no longer uses associated platform-specific error type and now has common type for `Error`
+- If the glyph you've requested is missing, the resulting missing glyph char will be in `Error::MissingGlyph`
```
chrisduerr, line 14:
```suggestion
- Missing glyphs are now returned as the content of the `MissingGlyph` error
```

CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- Rasterizer no longer uses associated platform-specific error type and now has common type for `Error`
```
chrisduerr, line 13:
```suggestion
- The rasterizer's `Error` type is now shared across platforms
```

src/darwin/mod.rs
```diff
@@ -144,20 +111,30 @@ impl crate::Rasterize for Rasterizer {
...
+        // Find a font where the given char is present.
+        let (font, glyph_index) = iter::once(font)
+            .chain(font.fallbacks.iter())
+            .find_map(|font| {
+                let glyph_index = font.glyph_index(glyph.character);
+                glyph_index.map(|glyph_index| (font, Some(glyph_index)))
+            })
+            .unwrap_or((font, None));
+
+        let is_missing_glyph = glyph_index.is_none();
```
chrisduerr, line 125: Can't we just check for `glyph_index == 0` later?

src/darwin/mod.rs
```diff
@@ -144,20 +111,30 @@ impl crate::Rasterize for Rasterizer {
...
+
+        // According to the index of 0 must be a missing glyph character:
```
chrisduerr, line 127:
```suggestion
        // According to the documentation, the index of 0 must be a missing glyph character:
```

src/directwrite/mod.rs
chrisduerr: I think the goal would be to see if missing glyph characters are displayed when a glyph isn't present and that normal glyphs aren't all suddenly broken.

More specific to this discussion: The question is if the notdef glyph has the same metrics as every other glyph of the font, so we could use it for the metrics instead of having to find the index for the `!` glyph.

src/directwrite/mod.rs
chrisduerr: With no other platform emitting this error, I feel like the best option is just to move this to the platform specific errors. There's no good way to handle this and it's just more likely to confuse anyone than to actually help.

I don't think this is something we've ever run into, so it's not unlikely that you'll only get to this case with weird setups like fonts that don't have the character specified or just not specifying any character at all.

src/directwrite/mod.rs
```diff
@@ -238,12 +231,30 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
-                let fallback_font = self.get_fallback_font(&loaded_font, glyph.c).ok_or(err)?;
-                self.rasterize_glyph(&fallback_font.create_font_face(), glyph.size, glyph.c)
-            },
-            result => result,
+        let fallback_font;
+        let (font, glyph_index) = iter::once(loaded_font)
+            .chain({
+                fallback_font =
+                    self.get_fallback_font(&loaded_font, glyph.character).map(Font::from);
+                &fallback_font
```
chrisduerr, line 239: At this point, why assign it in here when it's always done anyways? Might as well just declare it before the chain and just use `.chain(&fallback_font)`?

src/directwrite/mod.rs
```diff
@@ -238,12 +231,30 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
+            })
+            .find_map(|font| {
+                let glyph_index = self.get_glyph_index(&font.face, glyph.character);
+                glyph_index.map(|glyph_index| (font, Some(glyph_index)))
+            })
+            .unwrap_or((loaded_font, None));
+
+        // DirectWrite uses 0 for missing glyph symbols.
+        let is_missing_glyph = glyph_index.is_none();
```
chrisduerr, line 248: Same thing about just checking for `== 0`. If we can `.unwrap_or(0)`, we should also be able to check for `0`.

Can just create a constant too if we want to make it more clear.

src/directwrite/mod.rs
```diff
@@ -333,3 +322,10 @@ impl TextAnalysisSourceMethods for TextAnalysisSourceData<'_> {
         dwrite::DWRITE_READING_DIRECTION_LEFT_TO_RIGHT
     }
 }
+
+impl From<HRESULT> for Error {
+    fn from(hresult: HRESULT) -> Self {
+        let message = format!("A DirectWrite rendering error occurred: {:#X}", hresult);
```
chrisduerr, line 328: I really don't like the idea of using the formatter (`#`) here.

src/ft/mod.rs
```diff
@@ -399,15 +397,17 @@ impl FreeTypeRasterizer {
...
 
                     // We found something in a current face, so let's use it.
                     if index != 0 {
                         return Ok(font_key);
                     }
                 },
                 None => {
-                    if font_pattern.get_charset().map(|cs| cs.has_char(glyph.c)) != Some(true) {
+                    if font_pattern.get_charset().map(|cs| cs.has_char(glyph.character))
+                        != Some(true)
```
chrisduerr, line 409: Found a pretty good pattern to handle this in my selection PR the other day, I think this is far superior to both `.unwrap_or(false)` and `!= Some(true)`:

```suggestion
                    if font_pattern.get_charset().map_or(false, |cs| cs.has_char(glyph.character))
```

chrisduerr: Only works when you have `.map` already though unfortunately.

src/ft/mod.rs
```diff
@@ -427,7 +427,8 @@ impl FreeTypeRasterizer {
         // Render a normal character if it's not a cursor.
         let font_key = self.face_for_glyph(glyph_key)?;
         let face = &self.faces[&font_key];
-        let index = face.ft_face.get_char_index(glyph_key.c as usize);
+        let index = face.ft_face.get_char_index(glyph_key.character as usize);
+        let is_missing_glyph = index == 0;
```
chrisduerr, line 431: Index is never reassigned, right? So we should be able to just inline this check?

Also at this point I'm fairly certain we want a constant for this (for each platform, I think a global one is a bad idea even tough it seems to be the same on each platform).

src/ft/mod.rs
```diff
@@ -470,25 +471,30 @@ impl FreeTypeRasterizer {
...
-        if face.colored {
-            let fixup_factor = if let Some(pixelsize_fixup_factor) = face.pixelsize_fixup_factor {
-                pixelsize_fixup_factor
-            } else {
-                // Fallback if user has bitmap scaling disabled.
-                let metrics = face.ft_face.size_metrics().ok_or(Error::MissingSizeMetrics)?;
-                f64::from(pixelsize) / f64::from(metrics.y_ppem)
+        if is_missing_glyph {
+            Err(Error::MissingGlyph(rasterized_glyph))
+        } else if face.colored {
```
chrisduerr, line 488: What do you think about extracting the `face.colored` check? Fundamentally it just mutates the glyph and has nothing to do with returning something different.

I was thinking something like this maybe:

```rust
let mut rasterized_glyph = ...;

if face.colored {
    ...

    downsample_bitmap(&mut rasterized_glyph, fixup_factor);
}

if glyph_index == MISSING_GLYPH_INDEX {
    Err(Error::MissingGlyph(rasterized_glyph))
} else {
    Ok(rasterized_glyph)
}
```

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
     pub strikeout_thickness: f32,
 }
 
-pub trait Rasterize {
-    /// Errors occurring in Rasterize methods.
-    type Err: ::std::error::Error + Send + Sync + 'static;
+/// Errors occuring when using the rasterizer.
+#[derive(Debug)]
+pub enum Error {
+    /// Couldn't find font matching description.
```
chrisduerr, line 209:
```suggestion
    /// Unable to find a font matching the description.
```

`Couldn't` would work too, but I always feel like `Unable` just sounds a bit nicer? Either way the `a` and `the` would apply.

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
...
+    FontNotFound(FontDesc),
+
+    /// Tried to get size metrics from a Face that didn't have a size.
+    MissingSizeMetrics,
```
chrisduerr, line 213:
```suggestion
```

As per comment in the directwrite code.

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
...
+
+    MissingGlyph(RasterizedGlyph),
```
chrisduerr, line 215:
```suggestion
    /// The glyph could not be found in any font.
    MissingGlyph(RasterizedGlyph),
```

Alternatively:

```
/// Unable to find the glyph in any font.
```

But I personally like to mix things up a bit, instead of just always starting with the same words? Seems a bit redundant otherwise.

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
...
+        None
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::FontNotFound(font) => write!(f, "Unable to find the font: {:?}", font),
+            Error::MissingGlyph(glyph) => {
+                write!(f, "Unable to find glyph for char {}", glyph.character)
```
chrisduerr, line 235:
```suggestion
                write!(f, "no font with character {:?}", glyph.character)
```

Could go above and beyond and even print its value as bytes, but I think that's unnecessary.

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
...
+            },
+            Error::UnknownFontKey => f.write_str("Tried to use a font that hasn't been loaded"),
```
chrisduerr, line 237:
```suggestion
            Error::UnknownFontKey => f.write_str("invalid font key"),
```

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
...
+}
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```
chrisduerr, line 231: All errors should be lowercase and concise, based on what the Rust standard library is doing:

https://doc.rust-lang.org/src/std/io/error.rs.html#183-200

We're not formatting to the user here after all, these are descriptions of internal errors. Of course adding all relevant data (like character) is still useful.

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
...
+        match self {
+            Error::FontNotFound(font) => write!(f, "Unable to find the font: {:?}", font),
```
chrisduerr, line 233:
```suggestion
            Error::FontNotFound(font) => write!(f, "font {:?} not found", font),
```

src/lib.rs
```diff
@@ -203,23 +203,60 @@ pub struct Metrics {
...
+            Error::MissingGlyph(glyph) => {
+                write!(f, "Unable to find glyph for char {}", glyph.character)
+            },
+            Error::UnknownFontKey => f.write_str("Tried to use a font that hasn't been loaded"),
+            Error::MissingSizeMetrics => {
+                f.write_str("Tried to get size metrics from a face without a size")
+            },
+            Error::PlatformError(err) => write!(f, "{:?}", err),
```
chrisduerr, line 241:
```suggestion
            Error::PlatformError(err) => write!(f, "{}", err),
```

We don't want to quote the whole error, but instead just use the error text itself as error.

chrisduerr (changes requested):

src/lib.rs
kchibisov: I decided to go with slightly different formatting here.

src/directwrite/mod.rs
kchibisov: Oh, that's not intended, I'll revert it, since we shouldn't load a font unless we have missing glyph.

src/directwrite/mod.rs
kchibisov: >With no other platform emitting this error, I feel like the best option is just to move this to the platform specific errors. There's no good way to handle this and it's just more likely to confuse anyone than to actually help.

This is not true, we emit them from FreeType backend as well, platform reports them as None to us though....

kchibisov (commented):
I've updated things again, however I'm still not sure what do to about `MissingSizeMetrics`, since 2/3 platforms are using it, and FreeType backend doesn't provide any error for them, and returns just `None`... Should we just map it to some artificial error on FreeType backend?

src/directwrite/mod.rs
```diff
@@ -72,50 +75,42 @@ impl DirectWriteRasterizer {
...
-        let idx = *face
-            .get_glyph_indices(&[c as u32])
-            .first()
-            // DirectWrite returns 0 if the glyph does not exist in the font.
-            .filter(|glyph_index| **glyph_index != 0)
-            .ok_or_else(|| Error::MissingGlyph(c))?;
-
-        Ok(idx)
+    fn get_glyph_index(&self, face: &FontFace, character: char) -> u16 {
+        // DirectWrite returns 0 if the glyph does not exist in the font.
```
chrisduerr, line 102: This comment doesn't make a lot of sense anymore. If we would `unwrap` here, sure. But we make sure that everything works out and do `unwrap_or(0)` (which is good).

So if we talk about anything, we should probably talk about why we return `0` in the unwrap case? But I think we can just remove this comment completely.

Previously it made sense because we explicitly *filtered* the `0` out, but now we don't do that anymore and the constant makes it clear what the unwrap does. So there's no need to go into any more detail.

src/directwrite/mod.rs
```diff
@@ -238,12 +231,25 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
+        let mut glyph_index = self.get_glyph_index(&loaded_font.face, glyph.character);
+        let font = if glyph_index != MISSING_GLYPH_INDEX {
+            loaded_font
+        } else if let Some(font) = self.get_fallback_font(&loaded_font, glyph.character) {
+            fallback_font = Font::from(font);
+            glyph_index = self.get_glyph_index(&fallback_font.face, glyph.character);
+            &fallback_font
+        } else {
+            loaded_font
+        };
```
chrisduerr, line 244: This code looks a lot more complicated now because there's just way more branching one has to understand. What's the reason for this change? Could you maybe restructure it to use two conditionals at most?

One idea without changing too much code:

```rust
let mut font = &loaded_font;
let mut glyph_index = self.get_glyph_index(&loaded_font.face, glyph.character);
if glyph_index == MISSING_GLYPH_INDEX {
    if let Some(font) = self.get_fallback_font(&loaded_font, glyph_character) {
        glyph_index = self.get_glyph_index(&fallback_font.face, glyph.character);
        fallback_font = Font::from(font);
        font = &fallback_font;
    }
}
```

But I think we can do **way** better than that by using the abstraction magic of methods. Say we move the check for the glyph index in the `get_fallback_font` method. We could easily have an early return if the loaded font already contains the glyph, which would probably clean things up a lot. We also wouldn't need the `fallback_font` temporary mapping which is just annoying. I think lifetimes should allow for this (see my test: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7c84ae252dc58c23deaa6827e8072542).

src/ft/mod.rs
```diff
@@ -17,10 +17,12 @@ pub mod fc;
 use fc::{CharSet, FTFaceLocation, Pattern, PatternHash, PatternRef, Rgba};
 
 use super::{
-    BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, Rasterize, RasterizedGlyph, Size, Slant,
-    Style, Weight,
+    BitmapBuffer, Error, FontDesc, FontKey, GlyphKey, Metrics, Rasterize, RasterizedGlyph, Size,
+    Slant, Style, Weight,
 };
 
+const MISSING_GLYPH_INDEX: u32 = 0;
```
chrisduerr, line 24: Is this documented somewhere?

chrisduerr (changes requested):
> since 2/3 platforms are using it, and FreeType backend doesn't provide any error for them, and returns just None... Should we just map it to some artificial error on FreeType backend?

That's a good point actually, I noticed myself that FreeType was using it too and was considering if we should maybe still provide it.

I'm still not sure if this is useful to anyone, but if 2/3 platforms provide it we might as well give the option to handle it instead of just swallowing it. So the question is mostly about what to name it I suppose. Rather than calling it `MissingMetrics`, maybe just `MetricsNotFound`? That way we say that we couldn't find them, not that they don't exist. It might be a bit more accurate.

src/darwin/mod.rs
```diff
@@ -144,20 +115,24 @@ impl crate::Rasterize for Rasterizer {
...
-            // No fallback, give up.
-            Err(Error::MissingGlyph(glyph.c))
-        })
+        let font = self.fonts.get(&glyph.font_key).ok_or(Error::UnknownFontKey)?;
+
+        // Find a font where the given char is present.
+        let (font, glyph_index) = iter::once(font)
+            .chain(font.fallbacks.iter())
+            .find_map(|font| match font.glyph_index(glyph.character) {
+                MISSING_GLYPH_INDEX => None,
```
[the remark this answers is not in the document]
kchibisov: Idk, I've used such syntax for a long time.

src/directwrite/mod.rs
kchibisov: >But I think we can do way better than that by using the abstraction magic of methods. Say we move the check for the glyph index in the get_fallback_font method. We could easily have an early return if the loaded font already contains the glyph, which would probably clean things up a lot. We also wouldn't need the fallback_font temporary mapping which is just annoying. I think lifetimes should allow for this (see my test: 

I'm not sure about I understand your suggestion, since the fundamental problem here is that you create a new font, thus you don't have a reference to it at all. You can't return a reference to a stack local object from a method, like `get_fallback_font`, so this fallback should be stored **somewhere**, and I don't see where you'll store it with your approach.

src/ft/mod.rs
kchibisov: Well, yes, right on a method we're using to get a char index.

src/directwrite/mod.rs
```diff
@@ -238,12 +230,24 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
+        let loaded_fallback_font;
+        let mut font = loaded_font;
+        let mut glyph_index = self.get_glyph_index(&loaded_font.face, glyph.character);
+        if glyph_index == MISSING_GLYPH_INDEX {
+            if let Some(fallback_font) = self.get_fallback_font(&loaded_font, glyph.character) {
+                loaded_fallback_font = Font::from(fallback_font);
+                glyph_index = self.get_glyph_index(&loaded_fallback_font.face, glyph.character);
+                font = &loaded_fallback_font;
+            }
+        }
```
chrisduerr, line 242: > You can't return a reference to a stack local object from a method, like get_fallback_font, so this fallback should be stored somewhere, and I don't see where you'll store it with your approach.

Have you looked at my suggested code? It does exactly that.

src/ft/mod.rs
```diff
@@ -17,10 +17,14 @@ pub mod fc;
 use fc::{CharSet, FTFaceLocation, Pattern, PatternHash, PatternRef, Rgba};
 
 use super::{
-    BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, Rasterize, RasterizedGlyph, Size, Slant,
-    Style, Weight,
+    BitmapBuffer, Error, FontDesc, FontKey, GlyphKey, Metrics, Rasterize, RasterizedGlyph, Size,
+    Slant, Style, Weight,
 };
 
+/// FreeType uses 0 for missing glyph
```
chrisduerr, line 24:
```suggestion
/// FreeType uses 0 for the missing glyph.
```

src/lib.rs
```diff
@@ -203,23 +203,61 @@ pub struct Metrics {
...
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::FontNotFound(font) => write!(f, "font {:?} not found", font),
+            Error::MissingGlyph(glyph) => {
+                write!(f, "glyph for character {:?} not found", glyph.character)
+            },
+            Error::UnknownFontKey => f.write_str("invalid font key"),
+            Error::MetricsNotFont => {
+                f.write_str("unable to find metrics information on a font face")
```
chrisduerr, line 240:
```suggestion
                f.write_str("unable to find metrics for a font face")
```

chrisduerr (changes requested):

src/directwrite/mod.rs
kchibisov: I looked, and your example only works because of the fact that you create and return right in the end of method (I'm not even sure why rust allows that...). In our case we can't do that, I've tried and it fails with the correct reason for that
```
error[E0515]: cannot return value referencing temporary value
   --> src/directwrite/mod.rs:122:9
    |
122 |            fallback.map_characters(
    |   _________^
    |  |_________|
    | ||
123 | ||             &text_analysis_source,
124 | ||             0,
125 | ||             length,
...   ||
130 | ||             loaded_font.stretch,
131 | ||         ).mapped_font.map(Font::from).as_ref()
    | ||_____________________________________-________^ returns a value referencing data owned by the current function
    | |______________________________________|
    |                                        temporary value created here

```

Your example will also fail if you do `let m = Font; Some(&m)` for the exact reason.,

kchibisov: A more complete example

```rust
fn get_fallback_font<'a>(&'a self, loaded_font: &'a Font, character: char) -> Option<&'a Font> {
    if character == '\0' {
        return Some(loaded_font);
    }
    let fallback = self.fallback_sequence.as_ref()?;

    let mut buffer = [0u16; 2];
    character.encode_utf16(&mut buffer);

    let length = character.len_utf16() as u32;
    let utf16_codepoints = &buffer[..length as usize];

    let locale = get_current_locale();

    let text_analysis_source_data = TextAnalysisSourceData { locale: &locale, length };
    let text_analysis_source = TextAnalysisSource::from_text(
        Box::new(text_analysis_source_data),
        Cow::Borrowed(utf16_codepoints),
    );

    let fallback_result = fallback.map_characters(
        &text_analysis_source,
        0,
        length,
        &self.available_fonts,
        Some(&loaded_font.family_name),
        loaded_font.weight,
        loaded_font.style,
        loaded_font.stretch,
    ).mapped_font?;

    Some(&Font::from(fallback_result))
}
```

No, it doesn't work, because you refer to local data.

src/lib.rs
```diff
@@ -203,23 +203,59 @@ pub struct Metrics {
...
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::FontNotFound(font) => write!(f, "font {:?} not found", font),
+            Error::MissingGlyph(glyph) => {
+                write!(f, "glyph for character {:?} not found", glyph.character)
+            },
+            Error::UnknownFontKey => f.write_str("invalid font key"),
+            Error::MetricsNotFont => f.write_str("unable to find metrics for a font face"),
```
chrisduerr, line 239:
```suggestion
            Error::MetricsNotFont => f.write_str("metrics not found"),
```

chrisduerr (changes requested):
Just one tiny thing, after that everything should be good to go I think.

chrisduerr (approved):

davidhewitt:
I had a go at running alacritty master against this branch this evening (after a couple of naive renames to match the renamed fields here). Sorry for the delay.

I copied a bunch of awkward unicode glyphs from [this page](https://www.fileformat.info/info/unicode/block/cjk_unified_ideographs_extension_e/utf8test.htm). This is what I saw:

image

I get the same output with alacritty 0.5.0.

kchibisov:
>I had a go at running alacritty master against this branch this evening (after a couple of naive renames to match the renamed fields here). Sorry for the delay.

Have you patched alacritty to draw missing glyph from a new Error in renderer or you need a patch for that?

chrisduerr:
Also @davidhewitt if you're willing to test anyways, I'm **very** curious about changing the directwrite backend to use the glyph index 0 for font metrics.

<details>
<summary>Diff</summary>

```diff
diff --git a/src/directwrite/mod.rs b/src/directwrite/mod.rs
index 1166349..d49b864 100644
--- a/src/directwrite/mod.rs
+++ b/src/directwrite/mod.rs
@@ -163,11 +163,7 @@ impl crate::Rasterize for DirectWriteRasterizer {
 
         let line_height = f64::from(ascent - descent + line_gap);
 
-        // Since all monospace characters have the same width, we use `!` for horizontal metrics.
-        let character = '!';
-        let glyph_index = self.get_glyph_index(face, character);
-
-        let glyph_metrics = face.get_design_glyph_metrics(&[glyph_index], false);
+        let glyph_metrics = face.get_design_glyph_metrics(&[0], false);
         let hmetrics = glyph_metrics.first().ok_or(Error::MetricsNotFont)?;
 
         let average_advance = f64::from(hmetrics.advanceWidth) * f64::from(scale);
```
</details>

src/lib.rs
```diff
@@ -203,23 +203,59 @@ pub struct Metrics {
...
-    /// Errors occurring in Rasterize methods.
-    type Err: ::std::error::Error + Send + Sync + 'static;
+/// Errors occuring when using the rasterizer.
+#[derive(Debug)]
+pub enum Error {
+    /// Unable to find a font matching the description.
+    FontNotFound(FontDesc),
+
+    /// Unable to find metrics for a font face.
+    MetricsNotFont,
```
davidhewitt, line 213: Is this a possible typo of `MetricsNotFound` ?

davidhewitt (commented):
> Have you patched alacritty to draw missing glyph from a new Error in renderer or you need a patch for that?

Ah, yes I need a patch or alacritty branch I can pull from please 👍 

> Also @davidhewitt if you're willing to test anyways, I'm very curious about changing the directwrite backend to use the glyph index 0 for font metrics.

So I just gave this a very quick whirl, and it seems to work completely fine. Font size and positioning looked correct. It might be one of those changes where we should just go with it, and if users start reporting unexpected issues we can revert?

src/lib.rs
kchibisov: Thx.fixed.

kchibisov:
The patch for alacritty is as follows

```diff
diff --git a/alacritty/src/cursor.rs b/alacritty/src/cursor.rs
index 8c78218..f8586e2 100644
--- a/alacritty/src/cursor.rs
+++ b/alacritty/src/cursor.rs
@@ -41,12 +41,12 @@ pub fn get_underline_cursor_glyph(width: usize, line_width: usize) -> Rasterized
 
     // Create a custom glyph with the rectangle data attached to it.
     RasterizedGlyph {
-        c: ' ',
+        character: ' ',
         top: line_width as i32,
         left: 0,
         height: line_width as i32,
         width: width as i32,
-        buf: BitmapBuffer::RGB(buf),
+        buffer: BitmapBuffer::RGB(buf),
     }
 }
 
@@ -57,12 +57,12 @@ pub fn get_beam_cursor_glyph(height: usize, line_width: usize) -> RasterizedGlyp
 
     // Create a custom glyph with the rectangle data attached to it
     RasterizedGlyph {
-        c: ' ',
+        character: ' ',
         top: height as i32,
         left: 0,
         height: height as i32,
         width: line_width as i32,
-        buf: BitmapBuffer::RGB(buf),
+        buffer: BitmapBuffer::RGB(buf),
     }
 }
 
@@ -86,12 +86,12 @@ pub fn get_box_cursor_glyph(height: usize, width: usize, line_width: usize) -> R
 
     // Create a custom glyph with the rectangle data attached to it.
     RasterizedGlyph {
-        c: ' ',
+        character: ' ',
         top: height as i32,
         left: 0,
         height: height as i32,
         width: width as i32,
-        buf: BitmapBuffer::RGB(buf),
+        buffer: BitmapBuffer::RGB(buf),
     }
 }
 
@@ -102,11 +102,11 @@ pub fn get_block_cursor_glyph(height: usize, width: usize) -> RasterizedGlyph {
 
     // Create a custom glyph with the rectangle data attached to it.
     RasterizedGlyph {
-        c: ' ',
+        character: ' ',
         top: height as i32,
         left: 0,
         height: height as i32,
         width: width as i32,
-        buf: BitmapBuffer::RGB(buf),
+        buffer: BitmapBuffer::RGB(buf),
     }
 }
diff --git a/alacritty/src/renderer/mod.rs b/alacritty/src/renderer/mod.rs
index b347556..6018cd2 100644
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -10,8 +10,8 @@ use std::sync::mpsc;
 use std::time::Duration;
 
 use crossfont::{
-    BitmapBuffer, FontDesc, FontKey, GlyphKey, Rasterize, RasterizedGlyph, Rasterizer, Size, Slant,
-    Style, Weight,
+    BitmapBuffer, Error as RasterizerError, FontDesc, FontKey, GlyphKey, Rasterize,
+    RasterizedGlyph, Rasterizer, Size, Slant, Style, Weight,
 };
 use fnv::FnvHasher;
 use log::{error, info};
@@ -184,7 +184,7 @@ impl GlyphCache {
         // Need to load at least one glyph for the face before calling metrics.
         // The glyph requested here ('m' at the time of writing) has no special
         // meaning.
-        rasterizer.get_glyph(GlyphKey { font_key: regular, c: 'm', size: font.size })?;
+        rasterizer.get_glyph(GlyphKey { font_key: regular, character: 'm', size: font.size })?;
 
         let metrics = rasterizer.metrics(regular, font.size)?;
 
@@ -209,7 +209,7 @@ impl GlyphCache {
     fn load_glyphs_for_font<L: LoadGlyph>(&mut self, font: FontKey, loader: &mut L) {
         let size = self.font_size;
         for i in 32u8..=126u8 {
-            self.get(GlyphKey { font_key: font, c: i as char, size }, loader);
+            self.get(GlyphKey { font_key: font, character: i as char, size }, loader);
         }
     }
 
@@ -286,8 +286,10 @@ impl GlyphCache {
         let rasterizer = &mut self.rasterizer;
         let metrics = &self.metrics;
         self.cache.entry(glyph_key).or_insert_with(|| {
-            let mut rasterized =
-                rasterizer.get_glyph(glyph_key).unwrap_or_else(|_| Default::default());
+            let mut rasterized = rasterizer.get_glyph(glyph_key).unwrap_or_else(|err| match err {
+                RasterizerError::MissingGlyph(glyph) => glyph,
+                _ => Default::default(),
+            });
 
             rasterized.left += i32::from(glyph_offset.x);
             rasterized.top += i32::from(glyph_offset.y);
@@ -319,7 +321,11 @@ impl GlyphCache {
         let (regular, bold, italic, bold_italic) =
             Self::compute_font_keys(font, &mut self.rasterizer)?;
 
-        self.rasterizer.get_glyph(GlyphKey { font_key: regular, c: 'm', size: font.size })?;
+        self.rasterizer.get_glyph(GlyphKey {
+            font_key: regular,
+            character: 'm',
+            size: font.size,
+        })?;
         let metrics = self.rasterizer.metrics(regular, font.size)?;
 
         info!("Font size changed to {:?} with DPR of {}", font.size, dpr);
@@ -353,7 +359,7 @@ impl GlyphCache {
         let mut rasterizer = crossfont::Rasterizer::new(dpr as f32, font.use_thin_strokes())?;
         let regular_desc = GlyphCache::make_desc(&font.normal(), Slant::Normal, Weight::Normal);
         let regular = Self::load_regular_font(&mut rasterizer, &regular_desc, font.size)?;
-        rasterizer.get_glyph(GlyphKey { font_key: regular, c: 'm', size: font.size })?;
+        rasterizer.get_glyph(GlyphKey { font_key: regular, character: 'm', size: font.size })?;
 
         rasterizer.metrics(regular, font.size)
     }
@@ -1017,7 +1023,7 @@ impl<'a> RenderApi<'a> {
             c = ' ';
         }
 
-        let mut glyph_key = GlyphKey { font_key, size: glyph_cache.font_size, c };
+        let mut glyph_key = GlyphKey { font_key, size: glyph_cache.font_size, character: c };
 
         // Add cell to batch.
         let glyph = glyph_cache.get(glyph_key, self);
@@ -1026,7 +1032,7 @@ impl<'a> RenderApi<'a> {
         // Render visible zero-width characters.
         if let Some(zerowidth) = zerowidth.filter(|_| !hidden) {
             for c in zerowidth {
-                glyph_key.c = c;
+                glyph_key.character = c;
                 let mut glyph = *glyph_cache.get(glyph_key, self);
 
                 // The metrics of zero-width characters are based on rendering
@@ -1540,7 +1546,7 @@ impl Atlas {
             gl::BindTexture(gl::TEXTURE_2D, self.id);
 
             // Load data into OpenGL.
-            let (format, buf) = match &glyph.buf {
+            let (format, buf) = match &glyph.buffer {
                 BitmapBuffer::RGB(buf) => {
                     multicolor = false;
                     (gl::RGB, buf)
```

davidhewitt:
Aha nice, yes with that patch I see missing glyph markers:

image

chrisduerr (approved):

PR #22, opened 2020-12-29
directwrite: Honor global ClearType setting

r-c-f:
This only uses the DWRITE_TEXTURE_CLEARTYPE_3x1 texture type if
ClearType itself is enabled elsewhere on Windows, and configures the
rendering options accordingly. The winuser feature is enabled in winapi
to allow for the checks.

### DIFF
--- a/src/directwrite/mod.rs
+++ b/src/directwrite/mod.rs
@@ -1,5 +1,6 @@
 //! Rasterization powered by DirectWrite.
 
+use libc::c_void;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::ffi::OsString;
@@ -8,11 +9,14 @@ use std::os::windows::ffi::OsStringExt;
 use dwrote::{
     FontCollection, FontFace, FontFallback, FontStretch, FontStyle, FontWeight, GlyphOffset,
     GlyphRunAnalysis, TextAnalysisSource, TextAnalysisSourceMethods, DWRITE_GLYPH_RUN,
+    DWRITE_TEXTURE_TYPE,
 };
 
+use winapi::shared::minwindef::BOOL;
 use winapi::shared::ntdef::{HRESULT, LOCALE_NAME_MAX_LENGTH};
 use winapi::um::dwrite;
 use winapi::um::winnls::GetUserDefaultLocaleName;
+use winapi::um::winuser::{SystemParametersInfoA, SPI_GETFONTSMOOTHING};
 
 use super::{
     BitmapBuffer, Error, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style,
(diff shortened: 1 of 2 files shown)


src/directwrite/mod.rs
```diff
@@ -132,14 +143,61 @@ impl DirectWriteRasterizer {
...
+        if use_smoothing != 0 {
+            unsafe {
+                SystemParametersInfoA(
+                    SPI_GETFONTSMOOTHINGTYPE,
+                    0,
+                    &mut smoothing_type as *mut _ as *mut c_void,
+                    0,
+                );
+            }
+        }
```
chrisduerr, line 189: Why do we have to check twice here if font smoothing should be enabled or not? That seems rather strange.

src/directwrite/mod.rs
```diff
@@ -61,11 +69,16 @@ impl DirectWriteRasterizer {
...
-        );
+        let rendering_mode = match self.texture_type {
+            dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1 => face
+                .get_recommended_rendering_mode_default_params(
+                    em_size,
+                    self.device_pixel_ratio,
+                    dwrote::DWRITE_MEASURING_MODE_NATURAL,
+                ),
+            dwrote::DWRITE_TEXTURE_ALIASED_1x1 => dwrote::DWRITE_RENDERING_MODE_ALIASED,
+            _ => panic!("invalid texture type"),
```
chrisduerr, line 80: We shouldn't panic, instead the typesystem should be used to enforce that only the variants we want can be provided. One option would be an enum for example, or just a boolean really.

chrisduerr (changes requested):

r-c-f:
Looking into doing this a bit more properly, the double-check would make sense if the rendering mode were selectable between standard anti-aliasing and ClearType proper, but using proper enums exposes the fact that ClearType is not selectable -- DWRITE_RENDERING_MODE_NATURAL and DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL both evaluate to 4. As such I've reworded the commit to reflect what it's actually doing. 

Whether this is actually useful when done the right way is another matter, of course; the quality of rendering elsewhere suffers so dramatically with all anti-aliasing disabled that alacritty (using an outline approximation of a bitmap font, as DirectWrite doesn't support strikes in non-Asian typefaces) becomes in effect the one application I *don't* want using the system settings, leaving a local fork making it configurable the only sensible choice regardless of where this goes.

chrisduerr:
> Whether this is actually useful when done the right way is another matter, of course; the quality of rendering elsewhere suffers so dramatically with all anti-aliasing disabled that alacritty (using an outline approximation of a bitmap font, as DirectWrite doesn't support strikes in non-Asian typefaces) becomes in effect the one application I don't want using the system settings, leaving a local fork making it configurable the only sensible choice regardless of where this goes.

If this doesn't actually provide any benefit to you and no other user needs it, then it's probably best to not add any additional complexity that doesn't serve any purpose.

I'm curious if you've tried just always using `DWRITE_RENDERING_MODE_DEFAULT`? How does that behave, does it actually use the global setting or does it always use aliasing/noaliasing?

r-c-f:
I did try this initially, but it seems intended only as a placeholder (using it for rasterization yields yet another 80070057 error). Microsoft themselves seem to have acknowledged the deficiency here in the newer [`IDWriteFactory2::CreateGlyphRunAnalysis`](https://docs.microsoft.com/en-us/windows/win32/api/dwrite_2/nf-dwrite_2-idwritefactory2-createglyphrunanalysis) method with a parameter to set the antialiasing mode, but given a question of immediate usefulness and what would be prerequisite work in dwrote it's probably best to forget this indeed for the time being.

PR #25, opened 2021-05-25
Handle font face location errors

PabloMarcendo:
Fixes alacritty/alacritty#5118

The unwraps on line 53 of `src/ft/fc/pattern.rs` and line 412 of `src/ft/mod.rs` caused panics in the situation described in the issue.

### DIFF
--- a/src/ft/fc/pattern.rs
+++ b/src/ft/fc/pattern.rs
@@ -49,9 +49,11 @@ impl<'a> StringPropertyIter<'a> {
             //
             // Potential unsafety? What happens if the pattern is modified while this ptr is
             // borrowed out?
-            Some(unsafe {
-                mem::transmute(CStr::from_ptr(value as *const c_char).to_str().unwrap())
-            })
+            unsafe {
+                CStr::from_ptr(value as *const c_char)
+                    .to_str()
+                    .map_or(None, |x| Some(mem::transmute(x)))
+            }
         } else {
             None
         }
--- a/src/ft/fc/pattern.rs
+++ b/src/ft/fc/pattern.rs
@@ -49,11 +49,7 @@ impl<'a> StringPropertyIter<'a> {
             //
             // Potential unsafety? What happens if the pattern is modified while this ptr is
             // borrowed out?
-            unsafe {
-                CStr::from_ptr(value as *const c_char)
-                    .to_str()
-                    .map_or(None, |x| Some(mem::transmute(x)))
-            }
+            unsafe { mem::transmute(CStr::from_ptr(value as *const c_char).to_str().ok()?) }
         } else {
             None
         }
--- a/src/ft/mod.rs
+++ b/src/ft/mod.rs
@@ -25,7 +25,6 @@ use super::{
 /// https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_get_char_index
 const MISSING_GLYPH_INDEX: u32 = 0;
 
-#[derive(Clone)]
 struct FallbackFont {
     pattern: Pattern,
     key: FontKey,
@@ -83,9 +82,7 @@ impl fmt::Debug for FaceLoadingProperties {
 
 /// Rasterizes glyphs for a single font face.
 pub struct FreeTypeRasterizer {
-    library: Library,
-    faces: HashMap<FontKey, FaceLoadingProperties>,
-    ft_faces: HashMap<FtFaceLocation, Rc<FtFace>>,
+    free_type_loader: FreeTypeLoader,
     fallback_lists: HashMap<FontKey, FallbackList>,
     device_pixel_ratio: f32,
 }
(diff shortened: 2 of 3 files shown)


src/ft/mod.rs
```diff
@@ -409,9 +409,9 @@ impl FreeTypeRasterizer {
                     }
 
                     let pattern = font_pattern.clone();
-                    let key = self.face_from_pattern(&pattern, font_key)?.unwrap();
-
-                    return Ok(key);
+                    return self
+                        .face_from_pattern(&pattern, font_key)?
+                        .ok_or(Error::FontFaceLocationNotFound);
```
chrisduerr, line 414: I don't think we should return an error here. Doing so basically means that as soon as this font is encountered, no more fonts will be checked. We should just pretend that no glyph is inside this font by never returning its key. So just skipping to the next loop step should do the job.

chrisduerr (changes requested):
Thanks for taking a look at this.

Have you verified that this fixes the problem? I don't have access to macOS at the moment unfortunately.

src/ft/mod.rs
PabloMarcendo: You are right. When returning the error, alacritty showed a '▯'. I changed it and now alacritty simply uses another font, which is way better.

I did run into some trouble with the borrow checker, so I had to clone the list of fallback fonts.

src/ft/fc/pattern.rs
```diff
@@ -49,9 +49,11 @@ impl<'a> StringPropertyIter<'a> {
             //
             // Potential unsafety? What happens if the pattern is modified while this ptr is
             // borrowed out?
-            Some(unsafe {
-                mem::transmute(CStr::from_ptr(value as *const c_char).to_str().unwrap())
-            })
+            unsafe {
+                CStr::from_ptr(value as *const c_char)
+                    .to_str()
+                    .map_or(None, |x| Some(mem::transmute(x)))
```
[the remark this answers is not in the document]
PabloMarcendo: That looks simpler indeed. I changed it.

PabloMarcendo:
> Have you verified that this fixes the problem? I don't have access to macOS at the moment unfortunately.

I don't have access to macOS either, but the issue was reported on Arch Linux. I have verified that this fixes the problem on Arch Linux.

src/ft/mod.rs
```diff
@@ -390,7 +391,7 @@ impl FreeTypeRasterizer {
             return Ok(glyph.font_key);
         }
 
-        for fallback_font in &fallback_list.list {
+        for fallback_font in &fallback_list.list.clone() {
```
chrisduerr, line 394: Cloning is not an option here. This shouldn't be necessary.

chrisduerr (changes requested):

src/ft/mod.rs
PabloMarcendo: The problem was that the list over which this loop iterates is borrowed immutably from self, while in the loop the `face_from_pattern` call borrows self mutably. It used to be no problem, because the loop always ended after the call to `face_from_pattern'.

To fix it, I did some refactoring. I have inlined the `face_from_pattern` call. Because that causes code duplication, I moved as much of the non-mutating parts as possible to a new function. Since this new function basically just creates a new FaceLoadingProperties, I placed it in FaceLoadingProperties.

src/ft/mod.rs
```diff
@@ -409,9 +415,32 @@ impl FreeTypeRasterizer {
...
+                        let face = FaceLoadingProperties::new(&pattern, ft_face);
+
+                        debug!("Loaded Face {:?}", face);
+
+                        self.faces.insert(font_key, face);
+
+                        return Ok(font_key);
+                    } else {
+                        continue;
+                    }
```
chrisduerr, line 443: There is no reason for this additional code duplication. This is not a proper solution to work around the borrowchecking error

chrisduerr (changes requested):

PabloMarcendo:
I've made another attempt.

By borrowing individual fields, borrowing self is avoided and the borrow checker is satisfied.

chrisduerr (changes requested):
I don't think that this solution is as good as it can get. We basically have to modify several functions that are called from multiple places just to work around an issue that is completely unrelated to the function itself. So that's kinda awkward from a function API design perspective.

There are a couple of solutions to this I've considered, but a lot of them just introduce different pitfalls. I think the best way forward would be to basically go with what you've done already, but cleanup the function signatures by extracting library, faces and ft_faces into a separate struct.

If we define all the methods that do not need a fallback list on that struct instead, there shouldn't be any problems with "accidentally" capturing too much of it. But we can still keep using `self` everywhere whithout having to worry about trying to manually implement a partial move.

The most difficult thing is probably finding a reasonable name for this sub-struct that makes sense. But I'm open to suggestions here. Do you think you could get this working?

PabloMarcendo:
Naming things is hard.

chrisduerr (approved):
Beauty, LGTM.

PR #26, opened 2021-07-01
Add ability to reload font cache

### LINKED ISSUE #19: freetype/fontconfig: loading a font which was installed after initializing a library will soft 'fail'
If you install a font after initializing the rasterizer it won't be able to find that font.

I think we can reload fontconfig caches on running match one more time or something like that. Could add a function to reload fontconfig caches as well.

kchibisov:
Fixes #19.

---
This functions indeed solves the problem when calling from glyph reload path in alacritty's update display routine. I was trying to do all the calling part 'on-demand', but this functions is kind of slow(it's not that slow, but slow for startup) and slows down stratup time, that's why I've decided to add separate method to update font cache.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fontconfig not checking for new fonts installed after `Rasterizer` creation when calling `Rasterizer::load_font`.
+- Fontconfig not checking for fonts installed after `Rasterizer` creation
 
 ## 0.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- On systems using Fontconfig backend newly added fonts not visible upon `Rasterizer` recreation.
+- Fontconfig not checking for new fonts installed after `Rasterizer` creation when calling `Rasterizer::load_font`.
 
 ## 0.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- On systems using Fontconfig backend newly added fonts not visible upon `Rasterizer` recreation.
+
 ## 0.3.0
 
 ### Changed
--- a/src/ft/fc/mod.rs
+++ b/src/ft/fc/mod.rs
@@ -49,7 +49,7 @@ pub fn font_match(config: &ConfigRef, pattern: &PatternRef) -> Option<Pattern> {
 }
 
 /// Reloads the Fontconfig configuration files.
-pub fn bring_config_upto_date() {
+pub fn update_config() {
     unsafe {
         let _ = FcInitBringUptoDate();
     }
--- a/src/ft/fc/mod.rs
+++ b/src/ft/fc/mod.rs
@@ -5,6 +5,7 @@ use foreign_types::{ForeignType, ForeignTypeRef};
 
 use fontconfig::fontconfig as ffi;
 
+use ffi::FcInitBringUptoDate;
 use ffi::FcResultNoMatch;
 use ffi::{FcFontList, FcFontMatch, FcFontSort};
 use ffi::{FcMatchFont, FcMatchPattern, FcMatchScan};
@@ -44,6 +45,13 @@ pub fn font_match(config: &ConfigRef, pattern: &PatternRef) -> Option<Pattern> {
         } else {
             Some(Pattern::from_ptr(ptr))
         }
+    }
+}
+
+/// Reloads the Fontconfig configuration files.
+pub fn bring_config_upto_date() {
+    unsafe {
+        let _ = FcInitBringUptoDate();
     }
 }
 
--- a/src/ft/mod.rs
+++ b/src/ft/mod.rs
@@ -91,8 +91,8 @@ pub struct FreeTypeRasterizer {
     device_pixel_ratio: f32,
 
     /// Rasterizer creation time stamp to delay lazy font config updates
-    /// in `Rasterizer::laod_font`.
-    initialization_time_stamp: Option<Instant>,
+    /// in `Rasterizer::load_font`.
+    creation_timestamp: Option<Instant>,
 }
 
 #[inline]
@@ -111,7 +111,7 @@ impl Rasterize for FreeTypeRasterizer {
             loader: FreeTypeLoader::new()?,
             fallback_lists: HashMap::new(),
             device_pixel_ratio,
-            initialization_time_stamp: Some(Instant::now()),
+            creation_timestamp: Some(Instant::now()),
         })
     }
 
@@ -164,8 +164,8 @@ impl Rasterize for FreeTypeRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, size: Size) -> Result<FontKey, Error> {
-        if self.initialization_time_stamp.map_or(true, |t| t.elapsed() > RELOAD_DELAY) {
-            self.initialization_time_stamp = None;
+        if self.creation_timestamp.map_or(true, |t| t.elapsed() > RELOAD_DELAY) {
+            self.creation_timestamp = None;
             fc::update_config();
         }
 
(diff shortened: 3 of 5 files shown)


src/ft/mod.rs
```diff
@@ -155,6 +155,8 @@ impl Rasterize for FreeTypeRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, size: Size) -> Result<FontKey, Error> {
+        // Reload font files.
+        fc::reload_config();
```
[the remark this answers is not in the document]
kchibisov: The problem is that method takes like additional 50ms to startup time on my system. Providing additional parameter to 'not trigger this function pls' also isn't a solution. Maybe there's a way to call it on startup without overhead, but I'm not sure how (initialization is done, but it's still trying to bring things up to date).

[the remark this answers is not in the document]
chrisduerr: Ah, so the solution would be to remove this code I quoted and only have the method for reloading the cache, then call that on font (size) change?

Alternatively we could also probably try and store the age of our config and only reload if it's too old. Say for example one would load a thousand fonts, you wouldn't want to reload the config every time. So by storing a config timestamp and only updating when it's older than 'XYZ' that would probably get us most of the way there?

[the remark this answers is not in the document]
kchibisov: The thing is that this function already does that, so in continues call of FcInitBringUptoData it'll reload font only once, if you do five calls in a raw, the problem is the first call to this function.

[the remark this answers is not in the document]
chrisduerr: Oh okay, so this wouldn't be necessary for loading a ton of fonts, but it might still help with startup as an alternative at least.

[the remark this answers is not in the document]
kchibisov: >Alternatively we could also probably try and store the age of our config and only reload if it's too old. Say for example one would load a thousand fonts, you wouldn't want to reload the config every time. So by storing a config timestamp and only updating when it's older than 'XYZ' that would probably get us most of the way there?

I don't get it, if you download a thousand fonts and then do a config reload you expect to have all the fonts loaded(visible) to application. If you load this thousand font files one by one and update config after font update you expect a newly downloaded font be available right away. So I'm not sure what to you really want here.

The solution to make it work 'under the hood' is to allow ConfigReload only after X amount of times after startup, I think (will add a bunch of reads of 'time' though on startup).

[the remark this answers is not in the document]
chrisduerr: > I don't get it, if you download a thousand fonts and then do a config reload you expect to have all the fonts loaded(visible) to application.

My concern was not with downloading a thousand fonts online, but rather calling the `load_font` method a thousand times. As you've said that shouldn't be a problem if there's some sort of internal caching, but if there was none and you'd have to block for 50ms every time that would be a problem.

> The solution to make it work 'under the hood' is to allow ConfigReload only after X amount of times after startup, I think (will add a bunch of reads of 'time' though on startup).

That's basically accomplished by my suggestion of having an "age" associated with the actively loaded config. Since at startup the age would obviously be set to the startup time, thus the initial font load wouldn't trigger a config update.

src/ft/mod.rs
```diff
@@ -85,6 +89,12 @@ pub struct FreeTypeRasterizer {
     loader: FreeTypeLoader,
     fallback_lists: HashMap<FontKey, FallbackList>,
     device_pixel_ratio: f32,
+
+    /// Rasterizer creation time stamp to delay lazy font list updates
+    /// in `Rasterizer::laod_font`.
+    ///
+    /// if the value is `None`, update could be done unconditionally.
+    initialization_time_stamp: Option<Instant>,
```
chrisduerr, line 97:
```suggestion
    /// Rasterizer creation timestamp to delay lazy font config updates
    /// in `Rasterizer::load_font`.
    creation_timestamp: Option<Instant>,
```

src/ft/mod.rs
```diff
@@ -25,6 +26,9 @@ use super::{
 /// https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_get_char_index
 const MISSING_GLYPH_INDEX: u32 = 0;
 
+/// Font reload delay.
+const RELOAD_DELAY: Duration = Duration::from_secs(2);
```
chrisduerr, line 30:
```suggestion
/// Delay before font config reload after creating the rasterizer.
const CONFIG_RELOAD_DELAY: Duration = Duration::from_secs(2);
```

src/ft/fc/mod.rs
```diff
@@ -47,6 +48,13 @@ pub fn font_match(config: &ConfigRef, pattern: &PatternRef) -> Option<Pattern> {
     }
 }
 
+/// Reloads the Fontconfig configuration files.
+pub fn bring_config_upto_date() {
+    unsafe {
+        let _ = FcInitBringUptoDate();
+    }
+}
```
chrisduerr, line 56: Seems like you tried to mimic fontconfig's method name here? I personally find it absolutely atrocious tbh.

I'd much rather go with something like `update_config`, like you had it before.

chrisduerr (changes requested):
Looks good I'd say. This should provide a solid balance between ease of use and automatic config update.

I'm a bit skeptical about fontconfig caching the `FcInitBringUptoDate` internally, when it's not doing it at startup. But I suppose it *should* be fine if your testing showed that repeated calls are no problem.

src/ft/mod.rs
```diff
@@ -85,6 +89,10 @@ pub struct FreeTypeRasterizer {
     loader: FreeTypeLoader,
     fallback_lists: HashMap<FontKey, FallbackList>,
     device_pixel_ratio: f32,
+
+    /// Rasterizer creation time stamp to delay lazy font config updates
+    /// in `Rasterizer::laod_font`.
```
chrisduerr, line 94:
```suggestion
    /// Rasterizer creation time stamp to delay lazy font config updates
    /// in `Rasterizer::load_font`.
```

src/ft/mod.rs
```diff
@@ -85,6 +89,10 @@ pub struct FreeTypeRasterizer {
...
+    initialization_time_stamp: Option<Instant>,
```
chrisduerr, line 95:
```suggestion
    creation_timestamp: Option<Instant>,
```

I'd suggest creation here since it's shorter, but it should definitely be `timestamp` instead of `time_stamp`.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- On systems using Fontconfig backend newly added fonts not visible upon `Rasterizer` recreation.
```
chrisduerr, line 13: This isn't really accurate, is it? If things aren't visible upon `Rasterizer` (re)creation, this wouldn't help, since this only applies ***after*** the rasterizer has been created. So it applies to *loading* new fonts after rasterizer creation.

I think the following is a better description:

```
- Fontconfig not checking for new fonts installed after `Rasterizer` creation when calling `load_font`
```

src/ft/mod.rs
```diff
@@ -155,6 +164,11 @@ impl Rasterize for FreeTypeRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, size: Size) -> Result<FontKey, Error> {
+        if self.initialization_time_stamp.map_or(true, |t| t.elapsed() > RELOAD_DELAY) {
```
chrisduerr, line 167: Is there a reason why you decided to go without `.take()` here?

chrisduerr (changes requested):

src/ft/mod.rs
kchibisov: Yes, it works only once with `take`, so if you call this function twice before `RELOAD_DELAY` it'll try to reload font, since time stamp is `None`, however I don't want to make it `None` until the time elapsed.

chrisduerr: Oh, duh. I didn't actually think that one through.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fontconfig not checking for new fonts installed after `Rasterizer` creation when calling `Rasterizer::load_font`.
```
chrisduerr, line 13: This line is too long and has punctuation at the end.

src/ft/mod.rs
```diff
@@ -155,6 +164,11 @@ impl Rasterize for FreeTypeRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, size: Size) -> Result<FontKey, Error> {
+        if self.creation_timestamp.map_or(true, |t| t.elapsed() > RELOAD_DELAY) {
```
chrisduerr, line 167: If we don't use `take` here, I think we can fit in a `timestamp` for the variable in the closure without exceeding the max column count per line. That way it's a little nicer to read.

chrisduerr (changes requested):
Should be good to go after this.

chrisduerr (approved):

PR #27, opened 2021-07-07
get_family_names: idiomatic conversion

vbauerster:
Nothing special here, just a bit of idiomatic conversion with collect.


chrisduerr (approved):
Sure.

PR #28, opened 2021-08-01
Fix bitmap rendering with FreeType 2.11.0

kchibisov:
Calling `render_glyph` on bitmap fonts(emojis, etc) now returns error
with freetype 2.11, which wasn't the case before. This PR checks
whether the glyph is bitmap or not before doing any rendering.

Fixes https://github.com/alacritty/alacritty/issues/5353

### DIFF
--- a/src/ft/mod.rs
+++ b/src/ft/mod.rs
@@ -378,20 +378,22 @@ impl FreeTypeRasterizer {
         }
 
         // Transform glyphs with the matrix from Fontconfig. Primarily used to generate italics.
-        if let Some(matrix) = face.matrix.as_ref() {
-            let glyph = face.ft_face.raw().glyph;
-
-            unsafe {
+        unsafe {
+            let raw_glyph = face.ft_face.raw().glyph;
+            if let Some(matrix) = face.matrix.as_ref() {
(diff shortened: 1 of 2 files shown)


src/ft/mod.rs
```diff
@@ -378,20 +378,22 @@ impl FreeTypeRasterizer {
...
+                    let outline = &(*raw_glyph).outline;
 
                     freetype_sys::FT_Outline_Transform(outline, matrix);
                 }
             }
-        }
 
-        glyph.render_glyph(face.render_mode)?;
+            // Don't render bitmap glyphs, it results in error with freestype 2.11.0.
+            if (*raw_glyph).format != freetype_sys::FT_GLYPH_FORMAT_BITMAP {
```
chrisduerr, line 393: I'd assume this raw pointer dereference is the reason the unsafe scope had to be extended?

chrisduerr (approved):

src/ft/mod.rs
kchibisov: Yeah, and since we do that for other stuff I've extended `unsafe` block

WhyNotHugo:
Thanks!

PR #31, opened 2021-09-12
Adjust font fallback behavior on macOS

theodore-s-beers:
NB: I've been building Alacritty with my own fork of crossfont, and that's fine. I just thought of opening a PR in case my quasi-fix would be of use to other Mac users. (See [Alacritty issue 2800](https://github.com/alacritty/alacritty/issues/2800) for background.) If the PR isn't welcome, by all means close it.

Apart from comments, two changes have been made to the code. First, my testing has shown that Apple Symbols is already included in the cascade list being requested for Menlo Regular (English). So there's no need to add it back manually, as was done before. I wrote a test that prints the fallback list, which has around forty items (including what appear to be duplicates?). Yes, we lose ".Apple Symbols Fallback," but "Apple Symbols" itself is there.

Second, _if it's available_, we can add [Noto Sans](https://www.google.com/get/noto/) to the fallback list as another high-coverage font. I've verified that this does nothing if the font is not installed. To my knowledge, all of the characters that people have been complaining about (₽, ⸨ and ₙ come to mind) are provided in Noto Sans. With this added to the fallback list on an as-available basis, users who are seeing issues with uncommon characters could be advised to install a font. It could be a decent fix for a while. I have no idea what the path might be to getting macOS font fallback behavior in crossfont/Alacritty to be more like what we see in Terminal.app or iTerm2—where ultimately, afaict, every installed font will be checked for a glyph before giving up.

Thank you for your work on this library.

### DIFF
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -88,19 +88,17 @@ impl Descriptor {
                         .map(|desc| desc.to_font(size, false))
                         .collect::<Vec<_>>();
 
-                    // TODO, we can't use apple's proposed
-                    // .Apple Symbol Fallback (filtered out below),
-                    // but not having these makes us not able to render
-                    // many chars. We add the symbols back in.
-                    // Investigate if we can actually use the .-prefixed
-                    // fallbacks somehow.
-                    if let Some(descriptor) =
-                        descriptors_for_family("Apple Symbols").into_iter().next()
-                    {
-                        fallbacks.push(descriptor.to_font(size, false))
-                    };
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
I see no reason to remove one preinstalled standard font in favor of adding another non-standard custom Google font.

theodore-s-beers:
I'm confused: Apple Symbols wasn't being removed. It's already part of the fallback list. The end result was only to add Noto.

PR #32, opened 2021-10-25
Use appropriate style fallback fonts at Core Text

a5ob7r:
Previously crossfont using Core Text as backend has always returned regular style fallback fonts, even if the original font style is bold. This causes no correct rendering bold style glyph, especially for CJK characters. The reasons are below.

1. Causes font fallback frequently because many fonts have no CJK glyphs.
2. Lookups glyph at fallback fonts, but the fonts have only regular style.

We can't resolve No. 1 but instead can do No. 2. This commit returns appropriate style fallback fonts corresponding to the original font style.

This can check by a command below.

```bash
echo -e 'Hello, こんにちは, 你好, 안녕하세요\n\e[1mHello, こんにちは, 你好, 안녕하세요\e[0m'
```
These are screenshots on alacritty with this PR's branch.

Before

Screen Shot 2021-10-25 at 19 19 38

After

Screen Shot 2021-10-25 at 19 17 49

Note that this fix should affect even if italic or bold italic for CJK characters in theory, but not in fact. Maybe no italic or bold italic fonts are installed by default in macOS. This is the same even if terminal.app or iTerm2 by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Minimum Rust version has been bumped to 1.46.0
+- Core Text backend uses a current font as the original fallback font instead of Menlo
+
+### Fixed
+
+- Core Text backend ignoring style for font fallback
 
 ## 0.3.1
 
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -15,7 +15,7 @@ use core_graphics::font::{CGFont, CGGlyph};
 use core_graphics::geometry::{CGPoint, CGRect, CGSize};
 use core_text::font::{
     cascade_list_for_languages as ct_cascade_list_for_languages,
-    new_from_descriptor as ct_new_from_descriptor, CTFont,
+    new_from_descriptor as ct_new_from_descriptor, new_from_name, CTFont,
 };
 use core_text::font_collection::create_for_family;
 use core_text::font_collection::get_family_names as ct_get_family_names;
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Core Text backend has always returned regular style fallback fonts
```
chrisduerr, line 13:
```suggestion
- CoreText backend ignoring style for font fallback
```

I think this should better describe what actually got fixed and is desired.

chrisduerr (changes requested):
If I remember the previous implementation correctly, it seems like you've completely revamped your changes. Could you elaborate on why you've chosen the current patch as solution for the problem?

Seems like you're retrieving the fallback list for the active font, rather than always getting the fallback font for Menlo, which I'd generally say makes sense. Is that correct?

a5ob7r:
@chrisduerr Thanks for your comments.

> If I remember the previous implementation correctly, it seems like you've completely revamped your changes. Could you elaborate on why you've chosen the current patch as solution for the problem?

I have thought that no way to get a appropriate font style from the original font style name surely and correctly because the name may be not a standard name such as W6, Light, or Oblique. But `core_text::font::cascade_list_for_languages` returns fallback fonts which contain appropriate style font using CTFont directly. So I have wanted to get it using the method.

But, it was my misunderstanding. I found an easy way to do it and rewrite this patch.

> Seems like you're retrieving the fallback list for the active font, rather than always getting the fallback font for Menlo, which I'd generally say makes sense. Is that correct?

Yes. But I realize that the change to do it is out of this PR's purpose. So maybe I send PR to do it in the future, but now only change the fallback font style in this PR.

CHANGELOG.md
a5ob7r: OK. But I think that it is not `CoreText` but `Core Text`.

https://developer.apple.com/documentation/coretext?language=objc

chrisduerr:
> Yes. But I realize that the change to do it is out of this PR's purpose. So maybe I send PR to do it in the future, but now only change the fallback font style in this PR.

No worries, generally I think this code makes a lot more sense logically than what we've had previously. Unfortunately I didn't write the previous code so I'm not sure if there were any specific reasons for doing it this way, but your patch makes sense.

I'll fix the build in a separate PR and then hopefully this branch should just work.

CHANGELOG.md
chrisduerr: Good point, let's go with `Core Text` then.

chrisduerr: Good point, let's go with `Core Text` then.

chrisduerr (changes requested):

src/darwin/mod.rs
```diff
@@ -75,41 +75,31 @@ impl Descriptor {
...
+            // Use Menlo as an original font.
+            let font_name = match (
+                ct_font.symbolic_traits().is_bold(),
+                ct_font.symbolic_traits().is_italic(),
+            ) {
+                (true, true) => "Menlo-BoldItalic",
+                (true, _) => "Menlo-Bold",
+                (_, true) => "Menlo-Italic",
+                _ => "Menlo-Regular",
+            };
```
chrisduerr, line 87: Wait, did you change back to the old version again? Why?

chrisduerr (changes requested):

src/darwin/mod.rs
a5ob7r: As I have said previously, I think that changing the source fallback font from Menlo to a current font is out of this PR's purpose. The purpose is to change then only fallback font style. But as you said previously, it makes sense in general and I think so too. So I will send another PR to do it.

Should we do it in this PR? If so, I will do it again and add the change to changelog.

chrisduerr: > Should we do it in this PR? If so, I will do it again and add the change to changelog.

Honestly I don't mind either way, I'll just write the commit message to capture the purpose of the change accordingly.

But if you want to send a separate PR, I'd like to merge that first rather than merging a shitty version of this PR just to later fix it all up again.

a5ob7r: I got it. I'll change an original fallback font from Menlo-Regular to the current font again in this PR and write a commit message and CHANGELOG about the change.

a5ob7r: Done.

chrisduerr (approved):
LGTM, thanks for the PR.

PR #34, opened 2021-11-11
Add harfbuzz feature

fee1-dead:
This revives alacritty/alacritty#2677, adding a `shape` method to `FreeTypeRasterizer` when the feature flag is enabled. Ideally it would be on the `Rasterizer` trait, but I haven't learned how to add macos and windows support yet.

EDIT: note that the CI is not ran with the feature enabled at all. Let me know if I should change CI to also run harfbuzz feature enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ log = "0.4"
 [target.'cfg(not(any(target_os = "macos", windows)))'.dependencies]
 servo-fontconfig = "0.5.1"
 freetype-rs = "0.26"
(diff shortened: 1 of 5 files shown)


chrisduerr (changes requested):
What's the goal of adding shaping to a cross-platform **rasterization** library and not even having it available on any other platform? The approach here just doesn't make much sense to me.

fee1-dead:
Okay, looks like the best approach is to make some fields for FreeTypeRasterizer public to allow shaping to be done at alacritty's side. As well as changing storing characters to `GlyphId`s.

PR #40, opened 2022-04-06
Add proportional font support

chrisduerr:
Since crossfont was previously used solely by Alacritty, there was no
need for anything other than monospaced fonts. This patch adds minimal
proportional font support to the FreeType backend.

The most relevant part for proportional fonts is using the glyph's
individual advance rather than the font metric's advance. This is
exposed as a new `advance` field on `RasterizedGlyph`.

Besides advance, it's also possible for two rasterized glyphs to be
moved closer together by the font in a process called kerning. Crossfont
mimics the FreeType API here and provides the `kerning` method which
returns the relative offset of the second of two glyphs. Kerning for
OpenType fonts implemented in a `GPOS` table is not supported.

### DIFF
--- a/src/ft/mod.rs
+++ b/src/ft/mod.rs
@@ -173,7 +173,109 @@ impl Rasterize for FreeTypeRasterizer {
     }
 
     fn get_glyph(&mut self, glyph_key: GlyphKey) -> Result<RasterizedGlyph, Error> {
-        self.get_rendered_glyph(glyph_key)
+        // Render a normal character if it's not a cursor.
+        let font_key = self.face_for_glyph(glyph_key);
+        let face = &self.loader.faces[&font_key];
+        let index = face.ft_face.get_char_index(glyph_key.character as usize);
+        let pixelsize = face
+            .non_scalable
+            .unwrap_or_else(|| glyph_key.size.as_f32_pts() * self.device_pixel_ratio * 96. / 72.);
+
+        if !face.colored {
+            face.ft_face.set_char_size(to_freetype_26_6(pixelsize), 0, 0, 0)?;
+        }
+
+        unsafe {
+            let ft_lib = self.loader.library.raw();
+            freetype::ffi::FT_Library_SetLcdFilter(ft_lib, face.lcd_filter);
+        }
+
+        face.ft_face.load_glyph(index, face.load_flags)?;
+
+        let glyph = face.ft_face.glyph();
+
+        // Generate synthetic bold.
+        if face.embolden {
+            unsafe {
+                freetype_sys::FT_GlyphSlot_Embolden(glyph.raw()
+                    as *const freetype_sys::FT_GlyphSlotRec
+                    as *mut freetype_sys::FT_GlyphSlotRec);
+            }
+        }
+
+        // Transform glyphs with the matrix from Fontconfig. Primarily used to generate italics.
+        let advance = unsafe {
+            let raw_glyph = face.ft_face.raw().glyph;
+            if let Some(matrix) = face.matrix.as_ref() {
+                // Check that the glyph is a vectorial outline, not a bitmap.
+                if (*raw_glyph).format == freetype_sys::FT_GLYPH_FORMAT_OUTLINE {
+                    let outline = &(*raw_glyph).outline;
+
+                    freetype_sys::FT_Outline_Transform(outline, matrix);
+                }
+            }
+
+            // Don't render bitmap glyphs, it results in error with freestype 2.11.0.
+            if (*raw_glyph).format != freetype_sys::FT_GLYPH_FORMAT_BITMAP {
+                glyph.render_glyph(face.render_mode)?;
+            }
+
+            let advance = (*raw_glyph).advance;
+            (advance.x as i32 / 64, advance.y as i32 / 64)
+        };
+
+        let (pixel_height, pixel_width, buffer) =
+            Self::normalize_buffer(&glyph.bitmap(), &face.rgba)?;
+
+        let mut rasterized_glyph = RasterizedGlyph {
+            character: glyph_key.character,
+            top: glyph.bitmap_top(),
+            left: glyph.bitmap_left(),
+            width: pixel_width,
+            height: pixel_height,
+            advance,
+            buffer,
+        };
+
+        if index == MISSING_GLYPH_INDEX {
+            Err(Error::MissingGlyph(rasterized_glyph))
+        } else {
+            if face.colored {
+                let fixup_factor = match face.pixelsize_fixup_factor {
+                    Some(fixup_factor) => fixup_factor,
+                    None => {
+                        // Fallback if the user has bitmap scaling disabled.
+                        let metrics = face.ft_face.size_metrics().ok_or(Error::MetricsNotFound)?;
+                        f64::from(pixelsize) / f64::from(metrics.y_ppem)
+                    },
+                };
+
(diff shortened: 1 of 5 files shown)


src/ft/mod.rs
```diff
@@ -346,87 +448,6 @@ impl FreeTypeRasterizer {
         Ok(glyph.font_key)
     }
 
-    fn get_rendered_glyph(&mut self, glyph_key: GlyphKey) -> Result<RasterizedGlyph, Error> {
```
chrisduerr, line 349: There seems to be no reason for this to be a separate method, so I decided to just inline it.

src/ft/mod.rs
```diff
@@ -173,7 +173,109 @@ impl Rasterize for FreeTypeRasterizer {
...
+                };
+
+                rasterized_glyph = downsample_bitmap(rasterized_glyph, fixup_factor);
+            }
+
+            Ok(rasterized_glyph)
+        }
+    }
+
+    fn kerning(&mut self, left: GlyphKey, right: GlyphKey) -> (f32, f32) {
```
kchibisov, line 260: Hm, I don't really like that API, what do you think about accepting FontKey here as well? Since you basically load font and then start iterating over stuff? When it goes cross font there's no notion of kerning at all, since fonts are just different and we can assume zero if the glyph is missing in that font and load font.

src/ft/mod.rs
```diff
@@ -173,7 +173,109 @@ impl Rasterize for FreeTypeRasterizer {
...
+        // Generate synthetic bold.
+        if face.embolden {
+            unsafe {
+                freetype_sys::FT_GlyphSlot_Embolden(glyph.raw()
+                    as *const freetype_sys::FT_GlyphSlotRec
+                    as *mut freetype_sys::FT_GlyphSlotRec);
+            }
+        }
+
+        // Transform glyphs with the matrix from Fontconfig. Primarily used to generate italics.
```
kchibisov, line 206: This comment is now off. Probably should move closer to matrix and not to advance stuff?

Also, have you verified how advance work with emoji fonts and bitmaps? I'm worried that they could a little bit special here, so you'd need to scale advance for them.

kchibisov (changes requested):
You should also add a CHANGELOG entry for that, saying that it's linux only for now.

src/ft/mod.rs
kchibisov: Ah, sorry, I completely forgot how crossfont works here, it's fine, nvm me.

kchibisov:
@chrisduerr beep.

chrisduerr:
@kchibisov No worries, I haven't lost track of this yet. I've just been sick over the last week so things have gotten a bit delayed.

kchibisov:
>@kchibisov No worries, I haven't lost track of this yet. I've just been sick over the last week so things have gotten a bit delayed.

Take your time. I've thought that you've changed stuff, but forgot to push or something, since you went through review.

chrisduerr:
Yeah thanks for the reminder, better safe than sorry.

chrisduerr:
Emojis work properly now:
tmp

src/ft/mod.rs
```diff
@@ -173,7 +173,113 @@ impl Rasterize for FreeTypeRasterizer {
     }
 
     fn get_glyph(&mut self, glyph_key: GlyphKey) -> Result<RasterizedGlyph, Error> {
-        self.get_rendered_glyph(glyph_key)
+        // Render a normal character if it's not a cursor.
+        let font_key = self.face_for_glyph(glyph_key);
```
kchibisov, line 177: That might sound strange, but what cursor? I think this comment is left over from crossfont separation?

src/ft/mod.rs
```diff
@@ -173,7 +173,113 @@ impl Rasterize for FreeTypeRasterizer {
...
+
+                // Scale glyph advance.
+                rasterized_glyph.advance.0 = (advance.0 as f64 * fixup_factor).round() as i32;
+                rasterized_glyph.advance.1 = (advance.1 as f64 * fixup_factor).round() as i32;
+
+                rasterized_glyph = downsample_bitmap(rasterized_glyph, fixup_factor);
+            }
+
+            Ok(rasterized_glyph)
+        }
```
kchibisov, line 261: Would you mind refactoring it into a match while we're at it? This nested isn't that readable tbh. If you don't want to, we can go ahead.

src/ft/mod.rs
```diff
@@ -173,7 +173,113 @@ impl Rasterize for FreeTypeRasterizer {
...
+                }
+            }
+
+            // Don't render bitmap glyphs, it results in error with freestype 2.11.0.
+            if (*raw_glyph).format != freetype_sys::FT_GLYPH_FORMAT_BITMAP {
+                glyph.render_glyph(face.render_mode)?;
+            }
+
+            let advance = (*raw_glyph).advance;
+            (advance.x as i32 / 64, advance.y as i32 / 64)
```
kchibisov, line 224: That might sound out of place, but we do have this `/ 64` all over the place, and we have `to_freetype_26_6`, however we could add `from_freetype_26_6`. Since it's not that readable tbh and looks like some kind of magic division.

kchibisov (changes requested):
That looks good, just some minor style rant, not entirely related to your code.

src/ft/mod.rs
chrisduerr: What exactly do you want to refactor into a match?

I'd suggest to just early return in missing glyphs to reduce nesting?

src/ft/mod.rs
chrisduerr: I was actually looking for a `from_freetype_26_6` method when writing this. It would 100% make sense to add it.

src/ft/mod.rs
kchibisov: I was thinking about something like
```rust
match index {
    MISSING_GLYPH_INDEX => Err(Error::MissingGlyph(rasterized_glyph)),
    _ if face.colored => {
        let fixup_factor = match face.pixelsize_fixup_factor {
            Some(fixup_factor) => fixup_factor,
            None => {
                // Fallback if the user has bitmap scaling disabled.
                let metrics = face.ft_face.size_metrics().ok_or(Error::MetricsNotFound)?;
                f64::from(pixelsize) / f64::from(metrics.y_ppem)
            }
        };

        // Scale glyph advance.
        rasterized_glyph.advance.0 = (advance.0 as f64 * fixup_factor).round() as i32;
        rasterized_glyph.advance.1 = (advance.1 as f64 * fixup_factor).round() as i32;

        rasterized_glyph = downsample_bitmap(rasterized_glyph, fixup_factor);
        Ok(rasterized_glyph)
    }
    _ => Ok(rasterized_glyph),
}
```

But early returns are fine too, I think.

src/ft/mod.rs
```diff
@@ -105,6 +105,27 @@ fn to_fixedpoint_16_6(f: f64) -> c_long {
...
+    fn into_f32(self) -> f32 {
+        self
+    }
+}
+
+impl IntoF32 for i64 {
+    fn into_f32(self) -> f32 {
+        self as f32
+    }
+}
```
chrisduerr, line 127: This trait is necessary because `i64` doesn't implement `Into<f32>` since it's not a lossless conversion.

kchibisov: Yeah, that sounds fine.

kchibisov (approved):

PR #46, opened 2022-06-12
Don't load disabled fonts on macOS

kchibisov:
This commit fixes crash due to crossfont trying to load disabled font.
Before loading the font, we ensure now that it's enabled.

### DIFF
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -7,6 +7,8 @@ use std::path::PathBuf;
 use std::ptr;
 
 use core_foundation::array::{CFArray, CFIndex};
+use core_foundation::base::{CFType, ItemRef, TCFType};
+use core_foundation::number::{CFNumber, CFNumberRef};
 use core_foundation::string::CFString;
 use core_graphics::base::kCGImageAlphaPremultipliedFirst;
 use core_graphics::color_space::CGColorSpace;
@@ -19,8 +21,10 @@ use core_text::font::{
 };
 use core_text::font_collection::create_for_family;
 use core_text::font_collection::get_family_names as ct_get_family_names;
+use core_text::font_descriptor;
 use core_text::font_descriptor::kCTFontColorGlyphsTrait;
 use core_text::font_descriptor::kCTFontDefaultOrientation;
+use core_text::font_descriptor::kCTFontEnabledAttribute;
 use core_text::font_descriptor::kCTFontHorizontalOrientation;
 use core_text::font_descriptor::kCTFontVerticalOrientation;
 use core_text::font_descriptor::SymbolicTraitAccessors;
@@ -168,7 +172,7 @@ impl crate::Rasterize for Rasterizer {
         }
     }
 
-    fn kerning(&mut self, left: GlyphKey, right: GlyphKey) -> (f32, f32) {
+    fn kerning(&mut self, _left: GlyphKey, _right: GlyphKey) -> (f32, f32) {
         (0., 0.)
     }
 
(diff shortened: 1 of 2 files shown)


kchibisov:
@sjml could you test this? Though at least you'll get an idea how it could be done.

src/darwin/mod.rs
```diff
@@ -287,7 +289,20 @@ fn cascade_list_for_languages(ct_font: &CTFont, languages: &[String]) -> Vec<Des
...
+        .filter(|fontdesc| unsafe {
+            fontdesc
+                .attributes()
+                .find(kCTFontEnabledAttribute)
+                .and_then(|attr_val| attr_val.downcast::<CFNumber>())
+                .map(|num| match num.to_i32() {
+                    Some(num) => num != 0,
+                    _ => false,
+                })
+                .unwrap_or(true)
```
chrisduerr, line 302:
```suggestion
                .and_then(|attr_val| attr_val.downcast::<CFNumber>())
                .and_then(CFNumber::to_i32)
                .map_or(false, |num| num != 0)
```

Could also use `!= Some(0)` in the last line, but I think `map_or` is superior since we're already in a pretty deep iterator chain here.

Also the entire filter could probably be easily moved to a method, seems like something that would be pretty simple to name and document so it likely makes sense.

chrisduerr (changes requested):

sjml:
I can't build the current master branch of alacritty if I sync right to this PR; I think it needs to be based on v0.3.1. I'll try applying the changes directly. 

```
error[E0063]: missing field `advance` in initializer of `RasterizedGlyph`
  --> alacritty/src/renderer/text/builtin_font.rs:87:20
   |
87 |             return RasterizedGlyph {
   |                    ^^^^^^^^^^^^^^^ missing `advance`

error[E0063]: missing field `advance` in initializer of `RasterizedGlyph`
   --> alacritty/src/renderer/text/builtin_font.rs:482:5
    |
482 |     RasterizedGlyph { character, top, left: 0, height: height as i32, width: width as i32, buffer }
    |     ^^^^^^^^^^^^^^^ missing `advance`
```

sjml:
Still seeing the same crash, unfortunately. I'll try and figure out what's happening differently here as opposed to the one from my janky PR.

sjml:
It looks like the `.attributes()` call doesn't return a full set, but only the `NSFontNameAttribute`, so the find can never properly get at `kCTFontEnabledAttribute`. :(

sjml:
Note that the way core-text gets attributes it has to use the private `.0` member of the descriptor. https://github.com/servo/core-foundation-rs/blob/e51cf4186de61bcc0a51d3c2d796d1ccfd9e88a9/core-text/src/font_descriptor.rs#L203

kchibisov:
@sjml could you retest, it should the same as what you're doing in core_text.

sjml:
We have a winner!

src/darwin/mod.rs
```diff
@@ -287,7 +291,24 @@ fn cascade_list_for_languages(ct_font: &CTFont, languages: &[String]) -> Vec<Des
     let list = ct_cascade_list_for_languages(ct_font, &langarr);
 
     // Convert CFArray to Vec<Descriptor>.
-    list.into_iter().map(|fontdesc| Descriptor::new(fontdesc.clone())).collect()
+    list.into_iter().filter(is_enabled).map(|fontdesc| Descriptor::new(fontdesc.clone())).collect()
+}
+
+fn is_enabled(fontdesc: &ItemRef<'_, CTFontDescriptor>) -> bool {
```
chrisduerr, line 297:
```suggestion
/// Check if a font is enabled.
fn is_enabled(fontdesc: &ItemRef<'_, CTFontDescriptor>) -> bool {
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #47, opened 2022-07-05
support "thin strokes" behavior on macOS Big Sur and Monterey

### LINKED ISSUE #36: use_thin_strokes has no effect on macOS 12
This is the changes that make it work
https://github.com/alacritty/alacritty/issues/4616#issuecomment-860810609

without this change, **use_thin_strokes** does nothing on macOS 12

can you merge this? 

```diff
ben@BeniMac2020 crossfont % git diff
diff --git a/src/darwin/mod.rs b/src/darwin/mod.rs
index aded7a7..9a0f84f 100644
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -445,10 +445,13 @@ impl Font {
 
         if use_thin_strokes {
             cg_context.set_font_smoothing_style(16);
+            cg_context.set_allows_font_smoothing(false);
+            cg_context.set_should_smooth_fonts(false);
+        } else {
+            cg_context.set_allows_font_smoothing(true);
+            cg_context.set_should_smooth_fonts(true);
         }
 
-        cg_context.set_allows_font_smoothing(true);
-        cg_context.set_should_smooth_fonts(true);
         cg_context.set_allows_font_subpixel_quantization(true);
         cg_context.set_should_subpixel_quantize_fonts(true);
         cg_context.set_allows_font_subpixel_positioning(true);
```

chrisnc:
fixes #36

Remove the `use_thin_strokes` parameter from `Rasterize::new`, which
only did anything on macOS and only prior to Big Sur. Instead, we will
enable or disable "font smoothing" on macOS based on the
`AppleFontSmoothing` user default.

Subpixel anti-aliasing, which was previously controlled by the
`CGFontRenderingFontSmoothingDisabled` user default, is no longer
available, and the API which WebKit uses to read this setting now always
returns `true`, regardless of its value in the defaults dictionary. [1] [2]
The `set_font_smoothing` function of the `darwin` module, which set this
value in user defaults, has likewise been removed.

macOS itself supports values of `AppleFontSmoothing` from 0 to 3. If it
is unset the system behaves as though the value were 2. For `crossfont`,
we'll treat any value other than 0 to mean "smoothing enabled", as that
is the interface Core Graphics provides for this. A future improvement
may be to decipher what values of 1 and 3 mean for Core Graphics/Core
Text rendering.

These changes let users get the "thin strokes" behavior by setting
`AppleFontSmoothing` to 0 with:

`defaults write -g AppleFontSmoothing -int 0`

(Or replace `-g` with a specific domain the setting should apply to,
rather than the whole system.)

Use `once_cell::sync::Lazy` to read the user default once, rather than
in every call to `get_glyph`.

Refactor the Darwin module to remove dead code and only export `CoreTextRasterizer`.

[1]: https://bugs.webkit.org/show_bug.cgi?id=216588
[2]: https://github.com/alacritty/crossfont/pull/47#issuecomment-1174723896

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,26 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- On macOS, use the `AppleFontSmoothing` user default to decide whether fonts should be "smoothed"
+
+### Changed
+
+- Renamed `darwin::Rasterizer` to `darwin::CoreTextRasterizer`
+
+### Fixed
+
+- On macOS, `use_thin_strokes` and `set_font_smoothing` did not work since Big Sur
+
+### Removed
+
+- `use_thin_strokes` parameter from `Rasterize::new` trait method
+- `set_font_smoothing` from the `darwin` module
+- `get_family_names` from the `darwin` module
+
 ## 0.4.2
 
 ### Fixed
@@ -16,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix 32-bit build with FreeType/Fontconfig backend.
+- Fix 32-bit build with FreeType/Fontconfig backend
 
 ## 0.4.0
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,8 @@ core-foundation = "0.9"
 core-text = "19"
 core-graphics = "0.22"
 core-foundation-sys = "0.8"
+objc = "0.2.7"
+once_cell = "1.12"
 
 [target.'cfg(windows)'.dependencies]
 dwrote = { version = "0.11" }
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -1,10 +1,13 @@
 //! Font rendering based on CoreText.
 
-#![allow(improper_ctypes)]
 use std::collections::HashMap;
+use std::ffi::CStr;
 use std::iter;
 use std::path::PathBuf;
 use std::ptr;
+
+use cocoa::base::{id, nil};
+use cocoa::foundation::{NSString, NSUserDefaults};
 
 use core_foundation::array::{CFArray, CFIndex};
 use core_foundation::base::{CFType, ItemRef, TCFType};
@@ -13,27 +16,22 @@ use core_foundation::string::CFString;
 use core_graphics::base::kCGImageAlphaPremultipliedFirst;
 use core_graphics::color_space::CGColorSpace;
 use core_graphics::context::CGContext;
-use core_graphics::font::{CGFont, CGGlyph};
+use core_graphics::font::CGGlyph;
 use core_graphics::geometry::{CGPoint, CGRect, CGSize};
 use core_text::font::{
     cascade_list_for_languages as ct_cascade_list_for_languages,
     new_from_descriptor as ct_new_from_descriptor, new_from_name, CTFont,
 };
 use core_text::font_collection::create_for_family;
-use core_text::font_collection::get_family_names as ct_get_family_names;
-use core_text::font_descriptor;
-use core_text::font_descriptor::kCTFontColorGlyphsTrait;
-use core_text::font_descriptor::kCTFontDefaultOrientation;
-use core_text::font_descriptor::kCTFontEnabledAttribute;
-use core_text::font_descriptor::kCTFontHorizontalOrientation;
-use core_text::font_descriptor::kCTFontVerticalOrientation;
-use core_text::font_descriptor::SymbolicTraitAccessors;
-use core_text::font_descriptor::{CTFontDescriptor, CTFontOrientation};
-
-use cocoa::base::{id, nil, NO};
-use cocoa::foundation::{NSOperatingSystemVersion, NSProcessInfo, NSString, NSUserDefaults};
+use core_text::font_descriptor::{
+    self, kCTFontColorGlyphsTrait, kCTFontDefaultOrientation, kCTFontEnabledAttribute,
+    CTFontDescriptor, SymbolicTraitAccessors,
+};
 
 use log::{trace, warn};
+use objc::rc::autoreleasepool;
+use objc::{class, msg_send, sel, sel_impl};
+use once_cell::sync::Lazy;
 
 pub mod byte_order;
 use byte_order::kCGBitmapByteOrder32Host;
@@ -51,11 +49,8 @@ const MISSING_GLYPH_INDEX: u32 = 0;
 ///
 /// The descriptor provides data about a font and supports creating a font.
 #[derive(Debug)]
-pub struct Descriptor {
-    family_name: String,
-    font_name: String,
+struct Descriptor {
     style_name: String,
-    display_name: String,
     font_path: PathBuf,
 
     ct_descriptor: CTFontDescriptor,
@@ -64,19 +59,15 @@ pub struct Descriptor {
 impl Descriptor {
     fn new(desc: CTFontDescriptor) -> Descriptor {
         Descriptor {
-            family_name: desc.family_name(),
-            font_name: desc.font_name(),
             style_name: desc.style_name(),
-            display_name: desc.display_name(),
             font_path: desc.font_path().unwrap_or_else(PathBuf::new),
             ct_descriptor: desc,
         }
     }
 
     /// Create a Font from this descriptor.
-    pub fn to_font(&self, size: f64, load_fallbacks: bool) -> Font {
+    fn to_font(&self, size: f64, load_fallbacks: bool) -> Font {
         let ct_font = ct_new_from_descriptor(&self.ct_descriptor, size);
-        let cg_font = ct_font.copy_to_CGFont();
 
         let fallbacks = if load_fallbacks {
             // TODO fixme, hardcoded en for english.
@@ -93,11 +84,7 @@ impl Descriptor {
             // Investigate if we can actually use the .-prefixed
             // fallbacks somehow.
             if let Ok(apple_symbols) = new_from_name("Apple Symbols", size) {
-                fallbacks.push(Font {
-                    cg_font: apple_symbols.copy_to_CGFont(),
-                    ct_font: apple_symbols,
-                    fallbacks: Vec::new(),
-                })
+                fallbacks.push(Font { ct_font: apple_symbols, fallbacks: Vec::new() })
             };
 
             fallbacks
@@ -105,28 +92,22 @@ impl Descriptor {
             Vec::new()
         };
 
-        Font { ct_font, cg_font, fallbacks }
+        Font { ct_font, fallbacks }
     }
 }
 
-/// Rasterizer, the main type exported by this package.
+/// CoreTextRasterizer, the main type exported by this package.
 ///
 /// Given a fontdesc, can rasterize fonts.
-pub struct Rasterizer {
+pub struct CoreTextRasterizer {
     fonts: HashMap<FontKey, Font>,
     keys: HashMap<(FontDesc, Size), FontKey>,
     device_pixel_ratio: f32,
-    use_thin_strokes: bool,
 }
 
-impl crate::Rasterize for Rasterizer {
-    fn new(device_pixel_ratio: f32, use_thin_strokes: bool) -> Result<Rasterizer, Error> {
-        Ok(Rasterizer {
-            fonts: HashMap::new(),
-            keys: HashMap::new(),
-            device_pixel_ratio,
-            use_thin_strokes,
-        })
+impl crate::Rasterize for CoreTextRasterizer {
+    fn new(device_pixel_ratio: f32) -> Result<CoreTextRasterizer, Error> {
+        Ok(CoreTextRasterizer { fonts: HashMap::new(), keys: HashMap::new(), device_pixel_ratio })
     }
 
     /// Get metrics for font specified by FontKey.
@@ -163,7 +144,7 @@ impl crate::Rasterize for Rasterizer {
             })
             .unwrap_or((font, MISSING_GLYPH_INDEX));
 
-        let glyph = font.get_glyph(glyph.character, glyph_index, self.use_thin_strokes);
+        let glyph = font.get_glyph(glyph.character, glyph_index);
 
         if glyph_index == MISSING_GLYPH_INDEX {
             Err(Error::MissingGlyph(glyph))
@@ -181,7 +162,7 @@ impl crate::Rasterize for Rasterizer {
     }
 }
 
-impl Rasterizer {
+impl CoreTextRasterizer {
     fn get_specific_face(
         &mut self,
         desc: &FontDesc,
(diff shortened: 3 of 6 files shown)


chrisnc:
@kchibisov here is what I have.

I think the biggest gap here is how to handle AppleFontSmoothing being set to 1 or 3, which macOS does handle differently, but I don't see any public API for, or any documentation of what it is supposed to mean.

On the issue thread you said:

> Can you configure this default actually?

If you're asking whether this can be configured, I describe in more detail in the commit message and the code comments what we do with the setting, but let me know if I've misunderstood your question or haven't answered it there.

chrisnc:
Fixed to only use `cacao` and `lazy_static` in the macOS config.

chrisnc:
Another thing I discovered in my experiments is that `CGFontRenderingFontSmoothingDisabled` is fully deprecated now and doesn't do anything. macOS has a private API that WebKit uses for obtaining this: https://github.com/WebKit/webkit/blob/6e14685cafbd406eb0bf2a9dac5d3518213183a1/Source/WebCore/PAL/pal/spi/cg/CoreGraphicsSPI.h#L336

I tried on my machine running 12.4 to call this function to see what it does, and the disassembly shows it always returns true; it doesn't go and look up the corresponding user default that crossfont was using to control this.

CoreGraphics`CGFontRenderingGetFontSmoothingDisabled:
    0x188240784 <+0>: mov    w0, #0x1
    0x188240788 <+4>: ret

CoreGraphics`CGFontRenderingGetFontSmoothingDisabled:
    0x7ff81952d022 <+0>: pushq  %rbp
    0x7ff81952d023 <+1>: movq   %rsp, %rbp
    0x7ff81952d026 <+4>: movb   $0x1, %al
    0x7ff81952d028 <+6>: popq   %rbp
    0x7ff81952d029 <+7>: retq

Based on this WebKit issue https://bugs.webkit.org/show_bug.cgi?id=216588 this has likely been the case since Big Sur.

kchibisov:
How does it work? I think it doesn't do anything wrt the actual font-smoothing no? Since the Api to make font smooth or thin(like it was before) doesn't work anymore?

In particular would it allow us to achieve thin strokes, the ones with subpixel AA, which apple killed in all of their applications in macOS 10.12?

chrisnc:
As far as I know, subpixel AA is gone for good, but the "thin strokes" behavior, to my understanding, has meant to disable "font smoothing", which means telling macOS not to make your font stroke width artificially larger. The only setting that needs to change to control this is `set_should_smooth_fonts`.

Cargo.toml
```diff
@@ -25,11 +25,12 @@ freetype-rs = "0.26"
 pkg-config = "0.3"
 
 [target.'cfg(target_os = "macos")'.dependencies]
-cocoa = "0.24"
+cacao = "0.2.1"
 core-foundation = "0.9"
 core-text = "19"
 core-graphics = "0.22"
 core-foundation-sys = "0.8"
+lazy_static = "1.4.0"
```
kchibisov, line 33: Could you use `once_cell` instead?

Cargo.toml
```diff
@@ -25,11 +25,12 @@ freetype-rs = "0.26"
 pkg-config = "0.3"
 
 [target.'cfg(target_os = "macos")'.dependencies]
-cocoa = "0.24"
+cacao = "0.2.1"
```
kchibisov, line 28: Could we move without this dependency? Or is there a particular reason for it? What was the issue with e.g. `cocoa`? I'm skeptical since we have both of the in dependency graph in e.g. alacritty, since that's what being used by winit/glutin.

src/lib.rs
```diff
@@ -21,11 +20,10 @@ pub mod directwrite;
 #[cfg(windows)]
 pub use directwrite::DirectWriteRasterizer as Rasterizer;
 
-// If target is macos, reexport everything from darwin.
 #[cfg(target_os = "macos")]
-mod darwin;
+pub mod darwin;
 #[cfg(target_os = "macos")]
-pub use darwin::*;
+pub use darwin::CoreTextRasterizer as Rasterizer;
```
kchibisov, line 26: What's the reason for those changes?

kchibisov (changes requested):
Haven't tested the changes yet, but the approach looks good. I don't like the `use_thin_stroke` option, since it doesn't really make sense on per application level.

Cargo.toml
chrisnc: Sure. I actually did it with `once_cell` originally but thought `lazy_static` was more common. I can change it back.

Cargo.toml
chrisnc: cacao provides a safe interface to interact with user defaults, while cocoa requires unsafe. I can switch if it's a blocker, but personally I think cacao is worth it. Up to you.

src/lib.rs
chrisnc: There were several warnings related to unused code and missing `pub` declarations in the darwin module, and these were partly masked by the fact that everything in darwin is re-exported, while the other backends only export their version of `Rasterizer`. Now that the `set_font_smoothing` function is gone, we can do the same thing with the darwin backend and only export its Rasterizer. No other interfaces should be needed. I renamed it to match how the other backends have unique names for their Rasterizers.

chrisnc:
Thanks for the review! I will address the requested changes later today.

Cargo.toml
kchibisov: `once_cell` is the Api that will get eventually into `std`, so once it'll be there we can safely drop the dependency on external crates entirely.

Cargo.toml
chrisnc: So, I remember now the other reason I used cacao... The cocoa crate only implements setting and reading boolean user defaults, and the type of this value is an int. I agree that cacao should be avoided for now, having looked more closely at the state of each, but I may need to do something special here. https://docs.rs/cocoa/0.24.0/cocoa/foundation/trait.NSUserDefaults.html

Cargo.toml
chrisnc: Fixed.

src/darwin/mod.rs
```diff
@@ -334,18 +271,41 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
+// We will interpret anything other than 0 as enabling font smoothing, and 0 as disabling it. Note
+// that setting this default as a false boolean value *does not* disable font smoothing; it must be
+// an integer 0.
+static FONT_SMOOTHING_ENABLED: Lazy<bool> = Lazy::new(|| unsafe {
+    let font_smoothing_key = NSString::alloc(nil).init_str("AppleFontSmoothing");
+    let font_smoothing_value: id =
+        msg_send![id::standardUserDefaults(), objectForKey: font_smoothing_key];
+    if msg_send![font_smoothing_value, isKindOfClass: class!(NSNumber)] {
+        match CStr::from_ptr(msg_send![font_smoothing_value, objCType]).to_bytes() {
+            b"q" | b"l" | b"i" | b"s" => {
```
kchibisov, line 285: Could you add description wrt those magic strings?

src/darwin/mod.rs
```diff
@@ -334,18 +271,41 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
     out
 }
 
+// The AppleFontSmoothing user default controls font smoothing on macOS. By default it is unset,
+// and the system behaves as though it is set to 2, which means a medium level of font smoothing.
+// We will interpret anything other than 0 as enabling font smoothing, and 0 as disabling it. Note
+// that setting this default as a false boolean value *does not* disable font smoothing; it must be
+// an integer 0.
+static FONT_SMOOTHING_ENABLED: Lazy<bool> = Lazy::new(|| unsafe {
+    let font_smoothing_key = NSString::alloc(nil).init_str("AppleFontSmoothing");
```
kchibisov, line 280: Are you sure that this string will be deallocated? Could we use `autoreleasepool` for the closure?

src/darwin/mod.rs
```diff
@@ -334,18 +271,41 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
+    let font_smoothing_value: id =
+        msg_send![id::standardUserDefaults(), objectForKey: font_smoothing_key];
+    if msg_send![font_smoothing_value, isKindOfClass: class!(NSNumber)] {
+        match CStr::from_ptr(msg_send![font_smoothing_value, objCType]).to_bytes() {
+            b"q" | b"l" | b"i" | b"s" => {
+                let font_smoothing_integer: id = msg_send![font_smoothing_value, integerValue];
+                font_smoothing_integer as i64 != 0
+            },
+            // the value is an NSNumber but not an integer
```
kchibisov, line 289:
```suggestion
            // The value is an NSNumber but not an integer.
```

src/darwin/mod.rs
```diff
@@ -334,18 +271,41 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
+            _ => true,
+        }
+    } else {
+        // the key does not exist or the value is not an NSNumber
```
kchibisov, line 293:
```suggestion
        // The key does not exist or the value is not an NSNumber.
```

src/darwin/mod.rs
```diff
@@ -334,18 +271,41 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
+// The AppleFontSmoothing user default controls font smoothing on macOS. By default it is unset,
+// and the system behaves as though it is set to 2, which means a medium level of font smoothing.
+// We will interpret anything other than 0 as enabling font smoothing, and 0 as disabling it. Note
+// that setting this default as a false boolean value *does not* disable font smoothing; it must be
+// an integer 0.
+static FONT_SMOOTHING_ENABLED: Lazy<bool> = Lazy::new(|| unsafe {
+    let font_smoothing_key = NSString::alloc(nil).init_str("AppleFontSmoothing");
+    let font_smoothing_value: id =
+        msg_send![id::standardUserDefaults(), objectForKey: font_smoothing_key];
+    if msg_send![font_smoothing_value, isKindOfClass: class!(NSNumber)] {
```
kchibisov, line 283: Could you early return from here altering condition to `!msg_send`?

kchibisov (changes requested):

kchibisov:
Should also add a CHANGELOG entry wrt user settings and wrt breaking rasterizer Api.

src/darwin/mod.rs
chrisnc: Sure, I will figure out the right way to free it. I was following the previous `set_font_smoothing` function which, now that you mention it, appears to leak a string each time the function is called...

src/darwin/mod.rs
```diff
@@ -572,4 +516,9 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn get_font_smoothing_enabled() {
+        println!("font smoothing enabled: {}", *super::FONT_SMOOTHING_ENABLED);
```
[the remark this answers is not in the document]
chrisnc: It was very helpful during development to make sure I was using the `objc` stuff correctly, rather than a separate binary that just read this value. If you feel strongly I can remove it.

src/darwin/mod.rs
chrisnc: Refactored to use early returns, and check for nil from objCType as well. (We check that it's an NSNumber beforehand, so it will be a buggy NSNumber that returns nil from objCType, but just to be safe. CStr::from_ptr does not check for NULL after all.)

src/darwin/mod.rs
chrisnc: Using `autoreleasepool` now.

src/darwin/mod.rs
chrisnc: Done.

chrisnc:
Addressed most recent feedback, added changelog entry.

chrisnc:
Was a little overzealous in my refactor of the objc code and managed to break the key lookup. It's working now. Ready for further review. Thanks!

CHANGELOG.md
```diff
@@ -6,6 +6,26 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
...
 
+## Unreleased
+
+### Added
+
+- On macOS, read `AppleFontSmoothing` from user defaults to decide whether fonts should be "smoothed".
+
+### Changed
+
+- `darwin::Rasterizer` renamed to `darwin::CoreTextRasterizer`
```
kchibisov, line 17:
```suggestion
- Renamed `darwin::Rasterizer` to `darwin::CoreTextRasterizer` on macOS
```

CHANGELOG.md
```diff
@@ -6,6 +6,26 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
...
+
+### Fixed
+
+- Get thin strokes working on macOS Big Sur and Monterey, enabled with `AppleFontSmoothing` set to 0.
```
kchibisov, line 21:
```suggestion
- `AppleFontSmoothing` not being respected when smoothing fonts on macOS
```

kchibisov (changes requested):

CHANGELOG.md
chrisnc: "on macOS" is redundant here, no? It's renamed for everybody, after all, it's just a module that is only used on macOS.

CHANGELOG.md
chrisnc: I was struggling with how to phrase this, but I feel like this is redundant with what is in "Added". The thing which is fixed is that there's no way to use thin strokes currently, and now there will be. I'll try a different wording.

src/darwin/mod.rs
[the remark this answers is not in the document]
chrisnc: I'd say it tests that the font smoothing thunk does not panic or crash, which is a possibility given that it's a medium-sized `unsafe` block. I'll figure out a way to turn this into a proper test though, to make sure different cases of user defaults are handled correctly.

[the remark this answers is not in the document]
chrisnc: Never mind. This thunk is executed already in the existing test, which calls `get_glyph`. There's not much value beyond that, so I will remove.

CHANGELOG.md
chrisnc: Moving `Renamed` to the front.

CHANGELOG.md
```diff
@@ -6,6 +6,26 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- On macOS, read `AppleFontSmoothing` from user defaults to decide whether fonts should be "smoothed".
```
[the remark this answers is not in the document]
chrisnc: Done.

chrisnc:
Updated changelog and commit message. Removed extraneous test.

CHANGELOG.md
chrisnc: We could remove the "Fixed" part altogether. I "fixed" the broken thing by removing it and adding something else. What do you think?

kchibisov (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,26 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
...
+
+- On macOS, use the `AppleFontSmoothing` user default to decide whether fonts should be "smoothed"
+
+### Changed
+
+- Renamed `darwin::Rasterizer` to `darwin::CoreTextRasterizer`
+
+### Fixed
+
+- On macOS, `use_thin_strokes` and `set_font_smoothing` do nothing in Big Sur and Monterey
```
chrisduerr, line 21: Due to the present tense "do nothing" here, it's not entirely clear here if this is how it works now, or if this is what got fixed.

src/darwin/mod.rs
```diff
@@ -334,18 +270,49 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
+
+        // NSNumber's objCType method returns one of these strings depending on the size:
+        // q = quad (long long), l = long, i = int, s = short.
+        // If objCType returns a different specifier the value might be a bool or an unsigned int.
+        let int_specifiers: [&[u8]; 4] = [b"q", b"l", b"i", b"s"];
+        if !int_specifiers.contains(&CStr::from_ptr(num_type as *const i8).to_bytes()) {
+            return true;
+        }
+
+        let smoothing: id = msg_send![value, integerValue];
```
chrisduerr, line 300: Can you remind me what happens when `msg_send!` is called without previously checking for `integerValue`? Does that cause an immediate crash?

Because if it just returns `nil` or something, we don't need to be checking against an array to validate the type.

chrisduerr (changes requested):

CHANGELOG.md
chrisnc: Will clarify, thanks.

src/darwin/mod.rs
chrisnc: At this point in the code, `value` is confirmed to be a (subclass of) `NSNumber`, and it responded with a non-nil value to `objCType`. Given the possible objCType values I'm checking for, it should have a well-defined response to `integerValue`. I want to check for the `objCType` specifically because of what I mentioned in the comment, that macOS itself requires this value to be an integer to have any effect. `defaults write -g AppleFontSmoothing -bool NO` does not turn smoothing off for anything else, but in that case `integerValue` would 0, and we would disable font smoothing. For bools, `objCType` returns `"c"`, which isn't in the list, so we leave smoothing enabled regardless of the value.

chrisnc:
Comments and changelog updated. Ready again.

CHANGELOG.md
chrisnc: Updated.

src/darwin/mod.rs
chrisduerr: Ah interesting, so it would return `0`, that's certainly interesting.

chrisduerr (approved):
Thanks for the PR.

chrisnc:
Thanks for the thorough reviews! I'll work on updating the corresponding alacritty PR this weekend. https://github.com/alacritty/alacritty/pull/6186

chrisduerr:
Thanks for helping out with macOS, always appreciated.

cattyhouse:
thanks for the PR, by the way, we can disable font smoothing for Alacritty only by specifying a domain ( _ps: the domain was io.alacritty but it is now **org.alacritty**_), since disabling font smooth for the whole system is probably not what most people want.
 
````defaults write org.alacritty AppleFontSmoothing -int 0````

chrisnc:
@cattyhouse yes, that's mentioned in the commit on Alacritty here: https://github.com/alacritty/alacritty/commit/2a676dfad837d1784ed0911d314bc263804ef4ef
Alacritty's bundle domain changed after the crossfont change was merged, but before the alacritty change was merged. crossfont is used in some other projects too, and so I mentioned in this commit message to use the appropriate domain if you don't want the global behavior. Personally I think it makes everything look better, though. [This post is a good description of what the default actually does](https://tonsky.me/blog/monitors/). tl;dr: This setting is not really "smoothing" anything anymore, just making things slightly thicker. To get accurate font rendering, you must disable it.

PR #50, opened 2022-10-23
Fix undeline metrics on macOS

kchibisov:
We use baseline as an indicator for undeline metrics across the backends, however macOS was converting them to descent relative leading to incorrect values.

### DIFF
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -318,14 +318,14 @@ impl Font {
     fn metrics(&self) -> Metrics {
         let average_advance = self.glyph_advance('0');
 
-        let ascent = self.ct_font.ascent() as f64;
-        let descent = self.ct_font.descent() as f64;
-        let leading = self.ct_font.leading() as f64;
-        let line_height = (ascent + descent + leading + 0.5).floor();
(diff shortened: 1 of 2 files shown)


src/darwin/mod.rs
```diff
@@ -319,13 +319,13 @@ impl Font {
         let average_advance = self.glyph_advance('0');
 
         let ascent = self.ct_font.ascent() as f64;
-        let descent = self.ct_font.descent() as f64;
+        let descent = self.ct_font.descent().floor() as f64;
```
chrisduerr, line 322: Why did you floor here? Where is this coming from?

chrisduerr (changes requested):

src/darwin/mod.rs
kchibisov: I've looked that freetype always gives integers, but macOS provide float, and this float doesn't make any sense, how descend coordinate could be 4.5(it should be negative, but it seems we use some weird api)? For us it just useless, so I did like it looks in freetype backend. Without that we have underline being bleed into font, and with that floor it looks like in `Terminal.app` for the same font.

kchibisov:
This should work.

chrisduerr (approved):

PR #57, opened 2023-12-03
Rework `Size` handling

kchibisov:
--

There're likely better ways to do that, but it really sucks to deal with those sizes on hidpi, since you can't pick some of them.

We also can't use `f32` directly, since it's not `Hash`, so we could only do thins like that with fractions, etc.

### DIFF
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -102,12 +102,11 @@ impl Descriptor {
 pub struct CoreTextRasterizer {
     fonts: HashMap<FontKey, Font>,
     keys: HashMap<(FontDesc, Size), FontKey>,
-    device_pixel_ratio: f32,
 }
 
 impl crate::Rasterize for CoreTextRasterizer {
-    fn new(device_pixel_ratio: f32) -> Result<CoreTextRasterizer, Error> {
-        Ok(CoreTextRasterizer { fonts: HashMap::new(), keys: HashMap::new(), device_pixel_ratio })
+    fn new() -> Result<CoreTextRasterizer, Error> {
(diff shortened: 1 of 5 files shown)


kchibisov:
You can see example here https://github.com/alacritty/alacritty/pull/7410

src/darwin/mod.rs
```diff
@@ -118,7 +117,7 @@ impl crate::Rasterize for CoreTextRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, size: Size) -> Result<FontKey, Error> {
-        let scaled_size = Size::new(size.as_f32_pts() * self.device_pixel_ratio);
+        let scaled_size = Size::new(size.as_pt());
```
chrisduerr, line 120: Name is wrong, right? It's just size or ponit size, not scaled size now?

This is also wrong a couple more times in this file, might be worth a quick grep for `scale`.

src/ft/mod.rs
```diff
@@ -342,7 +334,7 @@ impl FreeTypeRasterizer {
     /// Load a font face according to `FontDesc`.
     fn get_face(&mut self, desc: &FontDesc, size: Size) -> Result<FontKey, Error> {
         // Adjust for DPR.
-        let size = f64::from(size.as_f32_pts() * self.device_pixel_ratio * 96. / 72.);
+        let size = f64::from(size.as_pt() * 96. / 72.);
```
chrisduerr, line 337: Isn't this just `as_px`?

chrisduerr (changes requested):

chrisduerr (approved):

PR #60, opened 2023-12-09
Switch `Size::from_px` constructor to `f32`

chrisduerr:
1f8e5db changed the `Size::as_px` method to return a floating point value instead of an integer. This patch ensures the constructor is consistent with that.


kchibisov (changes requested):
You need a changelog entry for this one.

kchibisov (approved):

PR #63, opened 2024-02-11
Fix font smoothing detection on macOS

### LINKED ISSUE #62: Turned off font smoothing is not detected correctly on macOS
I was trying to figure out why my font rendering looked incorrect in Alacritty and I finally traced it to the way i set the font smoothing:
```
defaults -currentHost write -g AppleFontSmoothing 0
```
I have been doing this for years, and it seems apple allows a string as well as an integer for this. 

While I have fixed my problem by storing the value as an integer it might be useful to fix the detection to work the same as in macOS?

Tested fix:
https://github.com/erikolofsson/crossfont/commit/2844bc8e4afe0fe9c8159c8fe511a39d9c77437a

erikolofsson:
Fixes #62

Add support for parsing smoothing option when `AppleFontSmoothing` is a `NSString`.


src/darwin/mod.rs
```diff
@@ -277,26 +277,30 @@ static FONT_SMOOTHING_ENABLED: Lazy<bool> = Lazy::new(|| {
...
-        // q = quad (long long), l = long, i = int, s = short.
-        // This is done to reject booleans, which are NSNumbers with an objCType of "c", but macOS
-        // does not treat them the same as an integer 0 or 1 for this setting, it just ignores it.
-        let int_specifiers: [&[u8]; 4] = [b"q", b"l", b"i", b"s"];
-        if !int_specifiers.contains(&CStr::from_ptr(num_type as *const i8).to_bytes()) {
-            return true;
+            let smoothing: id = msg_send![value, integerValue];
+            return smoothing as i64 != 0;
+        } else if msg_send![value, isKindOfClass: class!(NSString)] {
+            let smoothing: i64 = msg_send![value, intValue];
```
kchibisov, line 299: you should use `integerValue` and do basically the same as code above. with checking.

kchibisov (changes requested):

src/darwin/mod.rs
kchibisov: I just read the docs https://developer.apple.com/documentation/foundation/nsstring/1414988-intvalue?language=objc

Telling to use integeterValue. Nothing more.

erikolofsson: Sorry, I was confused by the previous implementation storing an `NSInteger` in an `id` so I thought it was a ref-counted type.

erikolofsson: Similarly the `num_type` is also using the wrong type, it should be:
```
 let num_type: *const c_char = msg_send![value, objCType];
if num_type.is_null() {
    return true;
}
```
Should I include fix in PR?

kchibisov: I guess. Also, once you fixed something you should re-request review, otherwise I'm likely to miss it.

kchibisov (approved):
I've rebased and fixed up some clippy issues. Should be good to go.

chrisduerr (approved):

PR #69, opened 2024-05-21
Use `objc2` for determining whether font smoothing is enabled

madsmtm:
`objc2` / `objc2-foundation` is much more type-safe than `cocoa`, and automatically ensures that memory management-rules are upheld.

See also https://github.com/alacritty/copypasta/pull/74.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- On macOS, use `objc2` when determining whether font smoothing is enabled.
+
 ## 0.8.0
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,6 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
-
-### Changed
-
-- On macOS, use `objc2` when determining whether font smoothing is enabled.
-
 ## 0.8.0
 
 ### Changed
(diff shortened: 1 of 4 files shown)


src/darwin/mod.rs
```diff
@@ -274,30 +270,40 @@ fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
-    autoreleasepool(|| unsafe {
-        let key = NSString::alloc(nil).init_str("AppleFontSmoothing");
-        let value: id = msg_send![id::standardUserDefaults(), objectForKey: key];
+    autoreleasepool(|_| {
+        let value = unsafe {
+            NSUserDefaults::standardUserDefaults().objectForKey(ns_string!("AppleFontSmoothing"))
+        };
+        let Some(value) = value else {
+            return true;
+        };
```
chrisduerr, line 279: Don't use let else, use a match statement please.

src/darwin/mod.rs
```diff
@@ -274,30 +270,40 @@ fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
-            if num_type.is_null() {
-                return true;
-            }
+        // SAFETY: The values in `NSUserDefaults` are always subclasses of
+        // `NSObject`.
+        let value: Retained<NSObject> = unsafe { Retained::cast(value) };
+
+        if value.is_kind_of::<NSNumber>() {
+            // SAFETY: Just checked that the value is a NSNumber
+            let value: Retained<NSNumber> = unsafe { Retained::cast(value) };
```
chrisduerr, line 287: Why is there no safe API for this? Seems like an oversight.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The sections should follow the order `Added`, `Changed`, `Fixed`, and `Removed`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- On macOS, use `objc2` when determining whether font smoothing is enabled.
```
chrisduerr, line 13: Should this actually change anything for consumers of this crate?

chrisduerr (changes requested):

src/darwin/mod.rs
madsmtm: There is some work in progress on it, see https://github.com/madsmtm/objc2/issues/559, it's difficult due having previously chosen to try to allow `&mut` references. Will likely require breaking changes, too, but it's something that I'm prioritizing.

CHANGELOG.md
madsmtm: Nope, I'm fine with not having a changelog entry

chrisduerr: Yeah seems very internal, I don't think a changelog entry is necessary.

PR #70, opened 2024-10-24
Update dependencies

chrisduerr:
This updates all dependencies to their latest major versions.


src/ft/mod.rs
```diff
@@ -663,7 +663,6 @@ impl FreeTypeLoader {
     fn new() -> Result<FreeTypeLoader, Error> {
         let library = Library::init()?;
 
-        #[cfg(ft_set_default_properties_available)]
```
kchibisov, line 666: should drop it from the build.rs then?

PR #71, opened 2024-10-24
Update dependencies

chrisduerr:


Cargo.toml
```diff
@@ -10,24 +10,24 @@ readme = "README.md"
...
 [dependencies]
 libc = "0.2"
 foreign-types = "0.5"
 log = "0.4"
 
 [target.'cfg(not(any(target_os = "macos", windows)))'.dependencies]
-yeslogic-fontconfig-sys = "5.0.0"
-freetype-rs = "0.36.0"
+yeslogic-fontconfig-sys = "6.0.0"
+freetype-rs = "0.37.0"
```
kchibisov, line 22: Could you revert this one, since otherwise it'll start bundling libpng and libz sometimes, which is really not great.

It has the unbundled feature, but it'll make it non compile on ubuntu 22 at all, which is probably not a great idea yet.

I'll try to address that upstream once I have time.

kchibisov (changes requested):

kchibisov (approved):

PR #72, opened 2025-01-13
Fix leak in fontconfig backend after searching for font

kchibisov:


kchibisov:
I believe this is https://github.com/alacritty/alacritty/issues/5438 . After running `valgrind --leak-check=full` it doesn't show anything interesting in alacritty anymore(all points to video driver or dlopened stuff).

kchibisov:
We can probably backport it and add to 0.15 while at it.

src/ft/fc/mod.rs
```diff
@@ -57,26 +57,33 @@ pub fn update_config() {
 
 /// List fonts by closeness to the pattern.
 pub fn font_sort(config: &ConfigRef, pattern: &PatternRef) -> Option<FontSet> {
-    unsafe {
-        // What is this result actually used for? Seems redundant with
-        // return type.
-        let mut result = FcResultNoMatch;
+    // What is this result actually used for? Seems redundant with
+    // return type.
+    let mut result = FcResultNoMatch;
```
chrisduerr, line 62: Can we just inline this and remove the stupid comment?

src/ft/mod.rs
```diff
@@ -451,8 +451,7 @@ impl FreeTypeRasterizer {
                     }
                 },
                 None => {
-                    if !font_pattern.get_charset().map_or(false, |cs| cs.has_char(glyph.character))
-                    {
+                    if font_pattern.get_charset().is_some_and(|cs| cs.has_char(glyph.character)) {
```
chrisduerr, line 454: You removed the leading negation without changing the logic here it seems? Previously `None` would have triggered the `continue`, but now `None` will not.

```suggestion
                    if !font_pattern.get_charset().is_some_and(|cs| cs.has_char(glyph.character)) {
```

chrisduerr (changes requested):

chrisduerr (commented):
LGTM, but I'll hold off with the approval until I have some time to test against Alacritty.

chrisduerr (approved):
This makes no notable difference for me as far as memory usage is concerned. However it should also be noted that I cannot recreate the severe memory leakage I was getting previously on the latest Alacritty master. Previously I could easily get it to multiple gigabytes by just holding `ctrl +`, with current master it took me some serious effort to get to just 500MiB.

I'd argue this does nothing to address https://github.com/alacritty/alacritty/issues/5438, but at the same time I would say that issue is far too vague and half the commenters are basing their complaints on numbers they do not understand.

kchibisov:
I mean, this does remove _some leak_. You could try to put `malloc_trim(0)` as I said before and see how memory usage is affected.

PR #74, opened 2025-05-12
Use more `objc2` framework crates

madsmtm:
Remove a few unnecessary dependencies, update `objc2`, and use more of the framework crates from that.

Motivation: The `core-foundation`, `core-text` and `core-graphics` crates are being deprecated, see https://github.com/servo/core-foundation-rs/issues/729.

It'd be nice if we could get a release after this.

### DIFF
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -24,8 +24,8 @@ use core_text::font_descriptor::{
     self, kCTFontColorGlyphsTrait, kCTFontDefaultOrientation, kCTFontEnabledAttribute,
     CTFontDescriptor, SymbolicTraitAccessors,
 };
-use objc2::rc::{autoreleasepool, Retained};
-use objc2_foundation::{ns_string, NSNumber, NSObject, NSObjectProtocol, NSString, NSUserDefaults};
+use objc2::rc::autoreleasepool;
+use objc2_foundation::{ns_string, NSNumber, NSUserDefaults};
 
 use log::{trace, warn};
 use once_cell::sync::Lazy;
@@ -280,14 +280,7 @@ static FONT_SMOOTHING_ENABLED: Lazy<bool> = Lazy::new(|| {
             None => return true,
         };
 
-        // SAFETY: The values in `NSUserDefaults` are always subclasses of
-        // `NSObject`.
-        let value: Retained<NSObject> = unsafe { Retained::cast(value) };
-
-        if value.is_kind_of::<NSNumber>() {
-            // SAFETY: Just checked that the value is a NSNumber
-            let value: Retained<NSNumber> = unsafe { Retained::cast(value) };
-
+        if let Some(value) = value.downcast_ref::<NSNumber>() {
             // NSNumber's objCType method returns one of these strings depending on the size:
             // q = quad (long long), l = long, i = int, s = short.
             // This is done to reject booleans, which are NSNumbers with an objCType of "c", but
@@ -302,10 +295,8 @@ static FONT_SMOOTHING_ENABLED: Lazy<bool> = Lazy::new(|| {
 
             let smoothing = value.integerValue();
             smoothing != 0
-        } else if value.is_kind_of::<NSString>() {
-            // SAFETY: Just checked that the value is a NSString
-            let value: Retained<NSString> = unsafe { Retained::cast(value) };
-            let smoothing = unsafe { value.integerValue() };
+        } else if let Ok(value) = value.downcast::<NSNumber>() {
+            let smoothing = value.integerValue();
             smoothing != 0
         } else {
             true
--- a/src/darwin/mod.rs
+++ b/src/darwin/mod.rs
@@ -1,37 +1,32 @@
 //! Font rendering based on CoreText.
 
 use std::collections::HashMap;
-use std::ffi::CStr;
+use std::ffi::{c_void, CStr};
 use std::iter;
 use std::path::PathBuf;
-use std::ptr;
+use std::ptr::{self, NonNull};
+use std::slice;
 
-use core_foundation::array::{CFArray, CFIndex};
-use core_foundation::base::{CFType, ItemRef, TCFType};
-use core_foundation::number::{CFNumber, CFNumberRef};
-use core_foundation::string::CFString;
-use core_graphics::base::kCGImageAlphaPremultipliedFirst;
-use core_graphics::color_space::CGColorSpace;
-use core_graphics::context::CGContext;
-use core_graphics::font::CGGlyph;
-use core_graphics::geometry::{CGPoint, CGRect, CGSize};
-use core_text::font::{
-    cascade_list_for_languages as ct_cascade_list_for_languages,
-    new_from_descriptor as ct_new_from_descriptor, new_from_name, CTFont,
-};
-use core_text::font_collection::create_for_family;
-use core_text::font_descriptor::{
-    self, kCTFontColorGlyphsTrait, kCTFontDefaultOrientation, kCTFontEnabledAttribute,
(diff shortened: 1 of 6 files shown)


src/darwin/mod.rs
```diff
@@ -50,21 +45,27 @@ struct Descriptor {
...
 
 impl Descriptor {
-    fn new(desc: CTFontDescriptor) -> Descriptor {
-        Descriptor {
-            style_name: desc.style_name(),
-            font_path: desc.font_path().unwrap_or_default(),
-            ct_descriptor: desc,
-        }
+    fn new(desc: &CTFontDescriptor) -> Descriptor {
+        let style_name = unsafe { desc.attribute(kCTFontStyleNameAttribute) }
```
chrisduerr, line 53: This is a rather unusual unsafe usage, I think I'd prefer it being assigned to its own variable over chaining off its scope block.

Also what exactly makes this unsafe? I'd assume it's due to the parameter potentially not existing? Might be a good idea to have a safety comment on it.

src/darwin/mod.rs
```diff
@@ -80,9 +81,9 @@ impl Descriptor {
             // many chars. We add the symbols back in.
             // Investigate if we can actually use the .-prefixed
             // fallbacks somehow.
-            if let Ok(apple_symbols) = new_from_name("Apple Symbols", size) {
-                fallbacks.push(Font { ct_font: apple_symbols, fallbacks: Vec::new() })
-            };
+            let name = CFString::from_static_str("Apple Symbols");
+            let apple_symbols = unsafe { CTFont::with_name(&name, size, ptr::null_mut()) };
+            fallbacks.push(Font { ct_font: apple_symbols, fallbacks: Vec::new() });
```
chrisduerr, line 86: What if the font doesn't exist? It looks like previously we were handling failure here.

src/darwin/mod.rs
```diff
@@ -210,34 +211,31 @@ impl CoreTextRasterizer {
...
+    let langarr = {
+        let tmp: Vec<_> = languages.iter().map(|language| CFString::from_str(language)).collect();
+        CFArray::from_retained_objects(&tmp)
     };
 
-    // CFArray of CTFontDescriptorRef (again).
-    let list = ct_cascade_list_for_languages(ct_font, &langarr);
+    let list =
+        unsafe { ct_font.default_cascade_list_for_languages(Some(langarr.as_opaque())) }.unwrap();
+    // CFArray of CTFontDescriptorRef.
```
chrisduerr, line 221: Not really necessary it's already obvious in the signature.

src/darwin/mod.rs
```diff
@@ -210,34 +211,31 @@ impl CoreTextRasterizer {
...
-        let attr_val =
-            font_descriptor::CTFontDescriptorCopyAttribute(descriptor, kCTFontEnabledAttribute);
-
-        if attr_val.is_null() {
-            return false;
-        }
+fn is_enabled(fontdesc: &CTFontDescriptor) -> bool {
+    let Some(attr_val) = (unsafe { fontdesc.attribute(kCTFontEnabledAttribute) }) else {
+        return false;
+    };
```
chrisduerr, line 232: No let else. Please use a match statement.

src/darwin/mod.rs
```diff
@@ -210,34 +211,31 @@ impl CoreTextRasterizer {
...
 
-        let attr_val = CFType::wrap_under_create_rule(attr_val);
-        let attr_val = CFNumber::wrap_under_get_rule(attr_val.as_CFTypeRef() as CFNumberRef);
+    let Ok(attr_val) = attr_val.downcast::<CFNumber>() else {
+        return false;
+    };
```
chrisduerr, line 236: Again, please don't use let else.

src/darwin/mod.rs
```diff
@@ -251,17 +249,39 @@ fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
         create_for_family("Menlo").expect("Menlo exists")
     });
 
-    // CFArray of CTFontDescriptorRef (i think).
-    let descriptors = ct_collection.get_descriptors();
+    let descriptors = unsafe { ct_collection.matching_font_descriptors() };
     if let Some(descriptors) = descriptors {
-        for descriptor in descriptors.iter() {
-            out.push(Descriptor::new(descriptor.clone()));
+        // CFArray of CTFontDescriptorRef (i think).
```
chrisduerr, line 254:
```suggestion
```

Was never really a good comment. I think? Really? :smile:

src/darwin/mod.rs
```diff
@@ -251,17 +249,39 @@ fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
+
+        for descriptor in descriptors {
+            out.push(Descriptor::new(&descriptor));
         }
     }
 
     out
 }
 
+pub fn create_for_family(family: &str) -> Option<CFRetained<CTFontCollection>> {
```
chrisduerr, line 266: A doc comment would go a long way on this function.

src/darwin/mod.rs
```diff
@@ -367,19 +378,25 @@ impl Font {
...
-            .get_bounding_rects_for_glyphs(kCTFontDefaultOrientation, &[glyph_index as CGGlyph]);
+        let glyphs = [glyph_index as CGGlyph];
+        let bounds = unsafe {
+            self.ct_font.bounding_rects_for_glyphs(
+                CTFontOrientation::Default,
+                NonNull::from(&glyphs[0]),
+                ptr::null_mut(),
+                1,
+            )
+        };
```
chrisduerr, line 399: Sorry but what's the point of creating an array just to immediately go and only retrieve its first element?

src/darwin/mod.rs
```diff
@@ -400,50 +417,65 @@ impl Font {
...
 
-        self.ct_font.draw_glyphs(
-            &[glyph_index as CGGlyph],
-            &[rasterization_origin],
-            cg_context.clone(),
-        );
+        unsafe {
+            self.ct_font.draw_glyphs(
+                NonNull::from(&[glyph_index as CGGlyph][0]),
+                NonNull::from(&[rasterization_origin][0]),
```
chrisduerr, line 465: Again, why are we creating arrays just to immediately access their only element?

src/ft/mod.rs
```diff
@@ -71,15 +71,18 @@ impl fmt::Debug for FaceLoadingProperties {
...
+                &match self.render_mode {
+                    freetype::RenderMode::Normal => "Normal",
+                    freetype::RenderMode::Light => "Light",
+                    freetype::RenderMode::Mono => "Mono",
+                    freetype::RenderMode::Lcd => "Lcd",
+                    freetype::RenderMode::LcdV => "LcdV",
+                    freetype::RenderMode::Max => "Max",
+                    freetype::RenderMode::Sdf => "Sdf",
+                },
+            )
```
chrisduerr, line 85: Did `cargo fmt` change this for you? It's certainly not an improvement and I'm not sure why it would have changed.

chrisduerr (changes requested):
Just out of curiosity, but did you write this patch entirely by hand?

src/darwin/mod.rs
ur4t: `CTFontCreateWithName` called by `CTFont::with_name` is actually returning `CTFont`, not `CTFont?` ([Apple's document here](https://developer.apple.com/documentation/coretext/ctfontcreatewithname(_:_:_:))).
In `objc2-core-text`, they are treated differently and `CTFont` is checked within `CTFont::with_name`. 
In `core-text`, `CTFont` and `CTFont?` are mapped to the same `CTFontRef`, which is misleading.

PR #83, opened 2026-07-18
Add colored emoji support on Windows (DirectWrite)

eman007:
## Summary

Adds colored emoji (COLR) rendering to the DirectWrite backend on Windows. Color glyphs are decomposed into their per-color layers via `IDWriteFactory2::TranslateColorGlyphRun` and composited into an RGBA bitmap; non-color glyphs continue through the existing ClearType greyscale path unchanged.

Before this change, emoji rendered as monochrome outline glyphs on Windows. They now render in full color.

## Changes

- Add a `ColorFactory` wrapping `IDWriteFactory2`, used to enumerate color glyph runs.
- Decompose color glyph runs and composite their per-color layers into an RGBA `BitmapBuffer`.
- Fall back to the existing greyscale path when a glyph has no color layers (`DWRITE_E_NOCOLOR`), and honor the foreground palette index (`0xFFFF`).
- Enable the `winapi` `dwrite`/`dwrite_1`/`dwrite_2`/`d2d1`/`dcommon`/`unknwnbase`/`winerror` features required for the color APIs.

## Testing

- `cargo check` builds clean on stable (Windows).
- Verified visually in Alacritty built against this crossfont: emoji render in color.
- No effect on non-Windows backends — the new code lives entirely in the existing `cfg(windows)` DirectWrite module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Colored emoji rendering on Windows (DirectWrite `TranslateColorGlyphRun`)
+
 ### Fixed
 
 - Ignore colored SVG fonts in fontconfig backend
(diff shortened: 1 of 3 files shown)


chrisduerr:
https://github.com/alacritty/alacritty/blob/master/CONTRIBUTING.md#llmai-contributions

